PR #67, opened 2020-06-05
Fix some strings and add spanish translation

sguinetti:
This pull request adds Spanish translation. First it establishes which lines are not translated and then the corresponding translation.


kasnder:
Wow. This is amazing. Thank you so much!

kasnder:
My Spanish isn't the best, but looks like good quality overall!

kasnder:
I've transferred your translation to here: https://crowdin.com/project/trackercontrol

Not sure if you still use the app, but if you do, you might want to head over there.

another-sapiens:
@kasnder I've continued with the translation, fixed and completed everything I saw on that site. 

I'll see if I can open up a pull request with the file.

kasnder:
I’ve seen that! Many thanks. Just added it to the code. See here: #180

sguinetti:
> I've transferred your translation to here: https://crowdin.com/project/trackercontrol
> 
> Not sure if you still use the app, but if you do, you might want to head over there.

No problem. Thanks for the notice. I've been trying with Crowdin with other apps and I would can translate in a future, because I'm not very available for now.

PR #130, opened 2020-10-14
Improve energy efficiency by applying Reduce Size Energy Pattern

anasofiagribeiro:
This improves the energy efficiency of tracker-control-android by applying the Reduce Size Energy Pattern for mobile applications.
The energy pattern was applied in DownloadTask.java, ServiceSinkHole.java, Util.java and Common.java. The general idea is when receiving data from an URLConnection or subclasses, if possible, receive the data compressed by GZIP scheme. In this specific case, when receiving data from the connection, urlConnection, connection and conn variable respectivly, request for the data to be compressed.


kasnder:
Thanks. This is great! I’ll merge this once I’ve checked it.

kasnder:
I've merged this.

PR #161, opened 2021-02-06
Change VPN interface subnet to prevent conflicts

### LINKED ISSUE #159: [BUG] TrackerControl breaks DNS resolution if connected to a 10.1.10.0/24 network
Edit: Looking at your code, it seems the VPN interface uses a 10.1.10.0/24 network. 

Can you please change it to something more random so it doesn't conflict if a user uses the same subnet e.g. 10.157.234.1/24

-----------------------
My device is connected to a wifi network and receives the following IP configuration via DHCP:
IP: 10.1.10.52/24
DNS resolver: 10.1.10.1

I'm using android 11 with calyx os.

If tracker control is enabled, all apps on my phone break as they cannot perform DNS resolution. e.g. searching is broken in firefox, Nothing shows in TC even if traffic log is enabled. However I can still access e.g. http://1.1.1.1 

On the otherhand, if I change my phone to use a static IP configuration with a public DNS resolver e.g. 1.1.1.1 or 8.8.8.8, everything works fine, even if TC is enabled. 

I have tried the following settings
- strict and non-strict mode (no difference)
- monitoring for a test app enabled and disabled (no difference)
- always on VPN is disabled
- private DNS is disabled
- battery optimisation for TC is off

metronidazole:
This changes the VPN interface's subnet from 10.1.10.1/24 to 10.213.213.1/24 to prevent conflicts for those who use trackercontrol on a network which uses 10.1.10.1/24

Fixes #159

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityLog.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityLog.java
@@ -123,7 +123,7 @@ public class ActivityLog extends AppCompatActivity implements SharedPreferences.
         lvLog.setAdapter(adapter);
 
         try {
-            vpn4 = InetAddress.getByName(prefs.getString("vpn4", "10.1.10.1"));
+            vpn4 = InetAddress.getByName(prefs.getString("vpn4", "10.213.213.1"));
             vpn6 = InetAddress.getByName(prefs.getString("vpn6", "fd00:1:fd00:1:fd00:1:fd00:1"));
         } catch (UnknownHostException ex) {
             Log.e(TAG, ex.toString() + "\n" + Log.getStackTraceString(ex));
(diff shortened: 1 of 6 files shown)


metronidazole:
Have tested this myself and it works.

kasnder:
Thanks ever so much! I'll test this.

kasnder:
As I understand it, this new IP address is also within the range of private IP addresses and shouldn't pose a problem.

metronidazole:
Not needed since there is already a built in feature to change the interface configuration #159#issuecomment-774548334

kasnder:
As you saw, there are some hard-coded IP addresses in the app code, so this might have cause issues. Thanks for checking.

Great to hear that it works for you, and sorry for the inconvenience.

PR #178, opened 2021-04-10
Spanish Translation

another-sapiens:
Okay, this is my first pull request ever, so you might wanna double-check the file just in case. 

Anyways, I hope this helps! 
I love this app and hope this brings the spanish speaking community a little bit more privacy.


kasnder:
Thanks! I think I’ve added this in #180. ::

PR #182, opened 2021-04-13
Fixed translation and Spanish Readme

another-sapiens:
Added the Spanish translation for the Readme file. Both the short and long description.
I don't know if it's also necessary to add the screenshot files into that folder, so I'll leave it up to you.

Additionally, there were a few errors in the previous translation (and there may be one or two more) so I've done my best to fix them.


another-sapiens:
[The conflict is because I've deleted and re-added the new translation file]

kasnder:
Amazing stuff--thanks! Just one question. You've added two files `strings.xml`, one in the correct place and one in the root directory. Which one is the corrent one?

another-sapiens:
Oops, forgot to delete the one in the root directory!
The new strings.xml file is the one in the right place, can you delete the other one?

kasnder:
I've now set up an integration with Crowdin. As part of this, I re-wrote the F-Droid full_description. So, I also had to remove this part of the readme. I've merged the other changes.

another-sapiens:
@kasnder Thank you, excellent work!
I've translated the new Readme file and created a pull request.

Issue #191, opened 2021-05-10
Language option


yetanotherfossman:
First of all, thanks for this software!

Looks like this app is translated into other languages, but there's no option for change it on settings. This would be very helpful because the app not always recognize the language correctly.

kasnder:
What language doesn’t work for you?

yetanotherfossman:
Yes, exactly.

another-sapiens:
Can confirm the language defaults to English, at least in the `es-AR` locale.

another-sapiens:
Jackpot! I believe it will only detect tte language if it coincides with the locale in the code, and not other locales for that particular language. Tested out the Spanish (Spain) locale and it worked like a charm.

Also, I should point out that my translation falls more into the `es-AR` category, but used `es-ES` as I didn't know the difference at the time.

yetanotherfossman:
@another-sapiens Can confirm, when I changed my system language to es-ES it worked :)

another-sapiens:
I just created a pull request with a possible fix. Fingers crossed!

kasnder:
Thanks for the pull request. As I explained there that "the issue is with the translation service CrowdIn. I'll try to change the configuration of this service so that languages will be correct again."

kasnder:
Ok, I've figured it out. The changes will go online in the next release.

https://github.com/OxfordHCC/tracker-control-android/commit/8688bb312cf378243b1c113eebc294d5e0958ce7

another-sapiens:
Oops, I thought it had to do with something else (it made sense in my head).

Thank you for the fix!

PR #215, opened 2021-06-23
Spanish for all regions

another-sapiens:
This is a possible fix for #191 - apparently, using `values-es-rES` results in Spanish strings only being used if the device's locale is "Spanish (Spain)". Normally, other Spanish locales would default to `values-es` (default Spanish when no region-specific strings are found). 
In this case, though, they resort to `values` (English) instead, because there's no `values-es` to use.

In theory, if this file exists, the issue should be fixed. 
So, I just copied the existing `values-es-rES` into another folder, named `values-es`.  
Haven't tested it yet, though. 

If it works, this could be useful for other languages.


kasnder:
Thanks! I can't accept this unfortunately, since the issue is with the translation service CrowdIn. I'll try to change the configuration of this service so that languages will be correct again.

another-sapiens:
@kasnder  Oops, I thought it had to do with something else (it made sense in my head).

Thank you for the fix!

PR #224, opened 2021-07-05
Update strings.xml

sgse:
Just a typo?
2020 or 2021?


kasnder:
It’s a typo. The reason why I haven’t changed is that it’d delete all translations of this.

sgse:
Ok, then I will close my PR.
BTW.
Thank you for this very good app :-)

PR #233, opened 2021-07-22
Add description to app icon button

PrivacyDragon:
Original issue: The app list shows an app icon followed by it's name and sometimes some information thing. The app icon is a button. If you click on it, you disable internet access for that application. The problem is that the button is not labeled. A blind user can not hear what it is and what it does.

Fixed by: adding a content description for the app icons.


kasnder:
This looks great, thanks!

PR #282, opened 2022-01-19
Code cleanup, updated readme.md

the-weird-aquarian:
- updated gradle
- code cleanup
- also removed unnecessary code corresponding to android versions less than min sdk
- flattened some nested layouts
- updated README.md

NOTE: some functions are never used anywhere, however, I have not removed them. Instead they have been commented out and added "// NEVER USED" before that. They can be removed if not needed or can be used if required in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,20 @@
 
 [![Crowdin](https://badges.crowdin.net/trackercontrol/localized.svg)](https://crowdin.com/project/trackercontrol) [![License: GPL v3](https://img.shields.io/badge/License-GPLv3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
 
-**Download directly [here](https://github.com/TrackerControl/tracker-control-android/releases/latest/download/TrackerControl-githubRelease-latest.apk), from [F-Droid](https://f-droid.org/packages/net.kollnig.missioncontrol.fdroid), or a feature-reduced version from [Google Play](https://play.google.com/store/apps/details?id=net.kollnig.missioncontrol.play).**
+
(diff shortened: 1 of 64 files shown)


kasnder:
Many thanks for the contribution -- I'll need some time to review this.

You're right to point out that there's quite a bit of unused code and concepts that could need some update. The reason that I've been careful so far about this was that part of TC's is forked from NetGuard, so this was trying to maintain compatibility with upstream changes. If I may not use all of your updates to this part of the code, this will be the reason.

Regardless of that, it's a good idea to make sure that the code base of TC is in good shape, and I'm super thankful for you looking into it.

the-weird-aquarian:
> You're right to point out that there's quite a bit of unused code and concepts that could need some update. The reason that I've been careful so far about this was that part of TC's is forked from NetGuard, so this was trying to maintain compatibility with upstream changes. If I may not use all of your updates to this part of the code, this will be the reason.

Now that you mention a part of TC is forked from NetGuard, I finally understand what all those native codes meant. I took a quick look at NetGuard's code and yes there does seem to be unused/outdated stuff there. Seems like I have to open a PR there too. :smile: 

P.S: there are lots of leaks too. Especially in all AsyncTasks and a context leak in one of the classes (If I remember correctly it was in DownloadTask.java). But seems like all of them are from NetGuard's code.

kasnder:
> P.S: there are lots of leaks too. Especially in all AsyncTasks and a context leak in one of the classes (If I remember correctly it was in DownloadTask.java). But seems like all of them are from NetGuard's code.

I must admit that I wasn't familiar with those -- will be more careful about it in the future.

kasnder:
I've now moved this to a different PR, so I can edit it.

PR #294, opened 2022-02-18
Some miscellaneous improvements

TacoTheDank:
### 1st commit
- Update gradle wrapper to 7.3.3 using git bash

### 2nd commit
- Fix miscellaneous gradle deprecations

### 3rd commit
- Deleted tabTrackersPosition, as it has no other use
- Use the created stateAdapter instance to return the current position instead of directly calling it from the other class

### 4th commit
- Rename the class to better fit its actual name (as it is no longer technically a PagerAdapter, but a StateAdapter)


kasnder:
I'll check this, and merge soon.

kasnder:
I've reviewed and merged this now. Thanks for your support!

PR #303, opened 2022-03-30
Fix Spanish short description

Tobertet:
I just read the short description and decided to make this small contribution.


kasnder:
Hello there! Thanks for the PR, this is greatly appreciated. If it's no problem for you, might you be able to add these translations in our translation platform at <https://crowdin.com/project/trackercontrol>. This would be of immense help. Otherwise, I'm more than happy to manually add these myself, but it's just a lot easier to have everything in one place.

Having said that, I'm sorry that you had to make so many changes in the first place. I had hoped that by now the translations would be in a better state. Am glad you brought this up!

Tobertet:
Hi @kasnder, I updated the short description in crowdin but I could not modify the other translations.

For some reason they won't appear.

kasnder:
> Hi @kasnder, I updated the short description in crowdin but I could not modify the other translations.
> 
> For some reason they won't appear.

Ok, thanks for looking into it! There's indeed some weird problem with the translation service that I don't really understand.. I've included these now.

PR #401, opened 2023-12-04
fix null pointer exception in Tracker.getName() / DNS name correctly shown in TC-Log output

derMart:
Hey, I noticed when using the adb log output setting that I am getting null pointer exceptions if an app is requesting an IP which is *not* a tracker, and as such the private name variable of the Tracker class is null.

Also, I noticed that the adb log output is not showing the DNS names of the IPs, but only the instance address of the wrapping Expiring class.

This pull request addresses to fix both problems.

cheers
derMart


kasnder:
thanks for this. i'll test it and then merge.

PR #409, opened 2024-01-14
Fix grammar error in msg_started

Hedreon:
I think it sounds better when it's written as "Looking for trackers" rather than "Checking trackers" in the context of a permanent silent notification.

You can also decide whether or not you want to keep the ellipsis at the end of `msg_started` or not.


kasnder:
good point!

PR #539, opened 2026-04-03
Fix lock deadlocks, production NPE, connection leaks, and log typo

kasnder:
- Wrap all lock.readLock/writeLock operations in try-finally blocks in
  ServiceSinkhole to prevent permanent deadlocks if exceptions occur
  while locks are held (e.g. during DB access in blockKnownTracker).
  Also fixes prepareHostsBlocked where unlock() ran unconditionally
  even when the lock was never acquired due to early return.
- Replace disabled assert + chained NPE in tracker logcat logging:
  assert is a no-op in production Android, and ipToHost.get(daddr)
  can return null, causing a crash when log_logcat is enabled.
- Fix copy-paste bug in ActivityMain.onActivityResult logging
  requestCode instead of resultCode.
- Add HttpURLConnection.disconnect() in finally blocks in Common.kt
  and Util.java to prevent socket leaks on repeated network calls.
- Move svg.renderToPicture() off the main thread in CountriesFragment;
  it was executing inside View.post() which runs on the UI thread.

https://claude.ai/code/session_011HzyMPo2gd34cnXG4EWLYj

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityMain.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityMain.java
@@ -612,7 +612,7 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
 
     @Override
     protected void onActivityResult(int requestCode, int resultCode, final Intent data) {
-        Log.i(TAG, "onActivityResult request=" + requestCode + " result=" + requestCode + " ok="
+        Log.i(TAG, "onActivityResult request=" + requestCode + " result=" + resultCode + " ok="
                 + (resultCode == RESULT_OK));
         Util.logExtras(data);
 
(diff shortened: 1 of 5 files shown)


kasnder:
Rebased/reconciled onto current `master` and force-pushed (was `f54a96c`, now `85900ef`).

Per the review of the original diff against post-gotatun master, this keeps only the still-valid fixes and re-implements the lock hardening against the current `ServiceSinkhole` bodies:

**Applied cleanly:** `onActivityResult` log typo (`requestCode`→`resultCode`); the TC-Log NPE guard (both `tracker` and `ipToHost.get(daddr)` can be null on a cache hit); `HttpURLConnection.disconnect()` leak fixes in `Util.getOrganization` and `Common.fetch` (plus `reader.use{}`); and moving `svg.renderToPicture()` off the UI thread in `CountriesFragment`.

**Re-implemented (diff no longer applied after the gotatun refactor):** try-finally lock hardening on `isAddressAllowed` (readLock — hot packet path, lock acquired outside the try, unlock in `finally`), `prepareUidIPFilters` and `prepareForwarding` (writeLock).

**Dropped:** the `startNative` else-branch hunk (that code no longer exists) and the `prepareHostsBlocked` change (master already fixes it with a `locked` boolean guard).

Now a clean master + 1 commit (5 files, +49/-18), no conflicts, `compileFdroidDebugJavaWithJavac` BUILD SUCCESSFUL.

PR #541, opened 2026-04-03
Remove unnecessary VPN rebuilds on network changes

kasnder:
## Summary

TrackerControl always runs in filter mode and doesn't differentiate blocking rules by network type (WiFi vs mobile), unlike NetGuard. This PR removes the unnecessary VPN rebuilds triggered by network changes, reducing VPN disruptions for users.

- **Stop rebuilding VPN on network-type changes**: connectivity receiver (WiFi/mobile switch), phone state listener (2G/3G/4G generation), AP state receiver (hotspot toggle), and interactive state receiver (screen on/off) no longer trigger `reload()`. These only mattered for NetGuard's per-network blocking rules which TC doesn't use.
- **Simplify `getAllowedRules()`**: Removed all wifi/metered/roaming/generation/screen filtering logic. In filter mode, `listAllowed` doesn't affect VPN routing (builder uses `rule.apply`) and `mapUidAllowed` is never read in `isAddressAllowed()`. Now returns all rules.
- **Remove `reload()` interactive optimization**: The screen-based rule check (`screen_wifi`/`screen_other`) was dead code — all 49 callers pass `interactive=false`.
- **Preserve Doze exit reload**: The reload on exiting Doze mode is kept as it serves as a health check for the native tunnel thread after network restrictions are lifted.
- **Remove unused `ExecutorService`**: Was only used by old `interactiveStateReceiver` reload logic.
- **Preserve `last_metered` WiFi override**: The metered state adjustment for WiFi is kept to avoid any edge-case behavioral change in `isLockedDown()`.

### What still triggers VPN rebuilds (intentionally kept)
- App install/uninstall (changes VPN routing)
- Settings changes (blocklist, blocking mode, etc.)
- Doze exit (tunnel health check)
- Explicit user actions (lockdown toggle, etc.)

## Test plan
- [ ] Verify VPN stays up and tracker blocking works after WiFi/mobile network switches
- [ ] Verify VPN stays up after screen on/off cycles
- [ ] Verify tracker blocking resumes correctly after device exits Doze mode
- [ ] Verify app install/uninstall still triggers VPN update
- [ ] Verify changing blocking mode or tracker settings still takes effect

https://claude.ai/code/session_01PuXevPG3gUKEXWHrgRMptV

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -2454,9 +2454,10 @@ public class ServiceSinkhole extends VpnService {
         @Override
         @TargetApi(Build.VERSION_CODES.M)
         public void onReceive(Context context, Intent intent) {
-            Log.i(TAG, "Received " + intent);
+            // TrackerControl: skip VPN rebuild on hotspot state changes.
+            // TC's tracker blocking is independent of tethering/AP state.
+            Log.i(TAG, "AP state changed (skipping reload): " + intent);
             Util.logExtras(intent);
-            reload("AP state changed", ServiceSinkhole.this, false);
         }
     };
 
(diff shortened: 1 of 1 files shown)


kasnder:
Closing as superseded. The network-change reload churn this PR targeted has since been addressed more surgically on master by `NetworkReloadPolicy` and `InteractiveStatePolicy` (debounced reloads), introduced with the gotatun/battery-fixes work merged in #594. This April branch now conflicts and would need to be re-derived against those classes rather than rebased. If a diff against the new policy classes shows a remaining reduction worth making, we can open a fresh, smaller PR. Thanks!

PR #556, opened 2026-04-05
Add option in filter settings to show froxzen (disabled) apps

### LINKED ISSUE #433: Checkbox to include frozen apps
### **Is your feature request related to a problem? Please describe.**

If you consider switching from another VPN firewall to TC on a particular system and give it a try, the process of migrating existing app rules to TC becomes a real nightmare if many frozen apps are involved. As of v2024.01.03, frozen apps don't show up in TC's app list at all - unless they get unfrozen. So, the workflow is:
1. install TC and check all three checkboxes in its apps list filter menu.
2. adjust all apps visible in TC's app list according to their respective settings in the other firewall
3. compare the apps list in TC and the other firewall
**4. for every app that is missing in TC,** 
**4.1**. find and open that app's details in system settings
**4.2**. unfreeze the app
**4.3**. scroll TC's app list all the way up (lacking a refresh button), pull down to refresh
**4.4**. scroll down to the position where the app was previously missing, find the app
**4.5**. adjust the app's rules according to its settings in the other firewall
**4.6**. switch back to that app's details in system settings
**4.7**. freeze the app again.

### **Describe the solution you'd like**

In TC's apps list filter menu, offer a fourth checkbox `Show frozen apps`, just below `Show apps without internet`.

### Additional Remarks

The problem is not limited to the described migration/test situation but occurs generally in TC with frozen apps: 
If you want to make sure the intended rules for a frozen app are already in place when it thaws, you need to be able to set those rules first, and therefor, make it visible in TC, and therefor, unfreeze it, and therefore .... you get it.

nmarteybotchway:
Closes #433

## Summary
Adds a new "Show frozen apps" filter checkbox to the main app list, 
allowing users to view and configure firewall rules for disabled/frozen 
apps without needing to unfreeze them first.

## Problem
Previously, disabled (frozen) apps were completely hidden from 
TrackerControl's app list. This made it impossible to:
- Pre-configure firewall rules before unfreezing an app
- Verify rules are in place for frozen apps
- Migrate rules from another firewall without temporarily unfreezing 
  every disabled app

## Changes
- `res/menu/main.xml`: Added `menu_app_frozen` checkbox item to the 
  filter submenu, below "Show apps without internet"
- `res/values/strings.xml`: Added `menu_show_frozen` string resource
- `ActivityMain.java` `onPrepareOptionsMenu`: Added checked state for 
  the new checkbox
- `ActivityMain.java` `onOptionsItemSelected`: Added handler that writes 
  `show_frozen` to SharedPreferences when toggled
- `ActivityMain.java` `onSharedPreferenceChanged`: Added `show_frozen` 
  to the preference change listener so the list refreshes automatically 
  when toggled
- `Rule.java` `getRules()`: Added `show_frozen` preference read and 
  updated the filter condition from `rule.enabled` to 
  `(show_frozen || rule.enabled)` so disabled apps are included when 
  the option is on

## Behaviour
- Toggle OFF (default): frozen/disabled apps hidden, existing behaviour 
  preserved
- Toggle ON: disabled apps appear in the list alongside active apps, 
  fully configurable

## Testing
- Disabled a system app via Settings → Apps → Disable
- Verified app was hidden with toggle off
- Verified app appeared in list with toggle on
- Verified firewall rules can be set on the disabled app
- Verified preference persists across app restarts
- Verified no visible change when no disabled apps are present on device

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityMain.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityMain.java
@@ -737,6 +737,7 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
                 "show_system".equals(name) ||
                 "show_nointernet".equals(name) ||
                 "show_unprotected".equals(name) ||
+                "show_frozen".equals(name) ||
                 "sort".equals(name) ||
                 "imported".equals(name)) {
             if ("sort".equals(name))
@@ -925,6 +926,7 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
 
         menu.findItem(R.id.menu_app_nointernet).setChecked(prefs.getBoolean("show_nointernet", true));
         menu.findItem(R.id.menu_app_unprotected).setChecked(prefs.getBoolean("show_unprotected", false));
+        menu.findItem(R.id.menu_app_frozen).setChecked(prefs.getBoolean("show_frozen", false));
 
         String sort = prefs.getString("sort", "trackers_week");
         if ("uid".equals(sort))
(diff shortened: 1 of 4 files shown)


kasnder:
Thanks!

PR #557, opened 2026-04-05
Add settings feature allowing to be walked through onboarding experience

nmarteybotchway:
Closes #553

## Summary
Adds a new "Restart introduction" preference to the settings screen, allowing users to replay onboarding experience at any time without needing to clear app data. 

## Changes
- `res/xml/preferences.xml`: Added new `category_help` preference category containing a `reintroduction` preference item with title and summary
- `res/values/strings.xml`: Added string resources for `setting_help`, `setting_reintroduction` and  `summary_reintroduction`
- `ActivitySettings.java` `configurePreferences()`: Added `findPreference("reintroduction")` click listener that launches`ActivityOnboarding` directly

## Behaviour
- New "Help" category appears in settings between the Backup and Further options sections
- Tapping "Restart introduction" launches the onboarding flow
- Onboarding completes normally and returns user to the app
- No preferences or data are reset - only the introduction is replayed

## Testing
- Tapped "Restart introduction" from settings
- Verified onboarding flow launches correctly
- Verified app returns to normal state after completing onboarding
- Verified app returns to normal state after dismissing onboarding
- Verified no data or settings are lost after replaying introduction


app/src/main/java/eu/faircode/netguard/ActivitySettings.java
```diff
@@ -740,7 +751,7 @@ else if ("vpn4".equals(name)) {
             ServiceSinkhole.reload("changed " + name, this, false);
 
         } else if ("validate".equals(name)) {
-            String host = prefs.getString(name, "www.f-droid.org");
+            String host = prefs.getString(name, "wpref_ww.f-droid.org");
```
kasnder, line 754: something is off here!

kasnder:
> Thanks @kasnder - I think the troubleshooting setting makes sense, and is quite convenient for users. You do raise a good point about most options from onboarding being available in the troubleshooting. To that end, do you think it would be better for us to include a new setting to change blocking level (dropdown or list of radio buttons)- that way there won't be a need for a "re-onboarding" section as the user will be able to configure all settings outside of the onboarding section. And maybe out of convenience, adding the restart of onboarding to the troubleshooting menu item for any users that want to quickly go through it again

But that already exists in the settings, no?

PR #585, opened 2026-05-13
feat: application insets

NYBACHOK:
I noticed that application UI overlaps with system bar, so added insets in this PR.


kasnder:
Superseded by #597, which fixes the same edge-to-edge overlap but follows the codebase's existing single-responsibility inset pattern (a narrowly-scoped listener on the bottom navigation) instead of a second root-level listener that returns `CONSUMED`. The root-`CONSUMED` approach competes with the existing `appbar` inset listener and swallows insets before they reach child views — the same double-inset bug previously fixed in `cd9887a`. Thanks for surfacing the issue and for the contribution!

PR #594, opened 2026-07-03
Migrate WireGuard engine from wireguard-go to gotatun (Rust)

kasnder:
## Summary

Replaces the Go `wgbridge` module (wireguard-go via gomobile) with a new Rust crate `wgbridge-rs/` embedding [gotatun](https://github.com/mullvad/gotatun) 0.7.2, Mullvad's WireGuard® implementation, and removes the Go toolchain from the repo and CI entirely.

**Motivation:** tunnel instability when roaming across networks/borders or under long uptime, higher battery drain, and reduced throughput. Mullvad hit the same wall — over 85% of their app crashes originated in wireguard-go, and after shipping gotatun to all Android users their user-perceived crash rate dropped from 0.40% to 0.01% ([announcement](https://mullvad.net/en/blog/announcing-gotatun-the-future-of-wireguard-at-mullvad-vpn)).

## What changed

- **`wgbridge-rs/`** — gotatun with custom transports for TrackerControl's unusual packet path: `IpRecv` over the socketpair written by the C hijack (`ip.c`), `IpSend` to the VpnService TUN fd with the passive DNS inspection ported from `dns.go` (test vectors included), and a `VpnService.protect()`-ed UDP factory that fails closed.
- **Hand-written Java bindings** in `net.kollnig.missioncontrol.wgbridge` with the same API shape as the gomobile-generated classes — `WgEgress`, the profile generators, and key rotation needed no call-site changes.
- **Roaming fixes** (the actual user-facing symptom): on network change or a broken-tunnel verdict, the bridge now rebinds the UDP sockets onto the new default network and re-resolves the peer endpoint (bypassing the 5-min cache) via new `Tunnel.rebind()` / `Tunnel.updateEndpoint()` — instead of tearing down and re-handshaking. Escalates to a full restart only if the tunnel breaks again within 90s.
- **Build/CI** — `wgbridgeBuild` Gradle task drives cargo-ndk for all four ABIs with 16KB-aligned LOAD segments (Android 15+); CI installs Rust instead of Go and runs `cargo test`. Contributor and F-Droid prerequisites documented in `wgbridge-rs/README.md`.

## Testing

- `cargo test`: 16/16 (DNS parser incl. compression pointers and IPv6 extension headers, UAPI config parsing, X25519 keys incl. RFC 7748 vector)
- `./gradlew testGithubDebugUnitTest`: passes, existing connectivity-monitor tests unchanged
- `./gradlew assembleGithubDebug`: APK bundles `libwgbridge.so` for all four ABIs (~0.8–1.3 MB each, vs the multi-MB Go runtime), zero Go artifacts, 16KB alignment verified with `llvm-readelf`
- ⚠️ **Not yet done:** on-device end-to-end testing (handshake, Wi-Fi↔cellular roaming recovery, DNS answers reaching the tracker mapping, multi-hour uptime)

## Notes

- gotatun is pinned to `=0.7.2`; its transport traits are the pre-1.0 surface most likely to change.
- gotatun has no direct "send keepalive now" API, so the monitor's prod pulses `persistent_keepalive` to 1s and restores it; rebind is triggered by a listen-port toggle that forces gotatun's connection re-setup through our protecting socket factory.


kasnder:
## On-device verification + follow-up fixes

Ran the migration end-to-end on an Android 15 emulator against a userspace WireGuard peer (no root/TUN on the host) that answers handshakes and serves HTTP/DNS *inside* the tunnel. This exercised the parts the PR flagged as "not yet done": handshake, data path through the C hijack → gotatun → UDP, passive DNS inspection landing in the tracker DB, fail-closed behaviour, and Wi-Fi roaming recovery.

**Result: works end to end.** TCP flowed through the tunnel (`hello-through-wireguard` rendered in Chrome), an in-tunnel DNS answer for `tracker-test.example` was recorded in the app's `dns` table (confirming the ported `dns.go` → `DnsRecorder` → `wireGuardDnsResolved` chain), roaming rebound the sockets and re-resolved the endpoint without tearing down, and with the peer killed the tunnel failed closed (no traffic escaped to the raw network) and auto-recovered when the peer returned. No crashes across the session.

Three rough edges turned up while verifying, now fixed in c50fdae:

**1. Status showed "Connected" before any handshake.** WireGuard handshakes lazily on first traffic, so the engine being up isn't the same as a working tunnel — a wrong endpoint read "Connected" until traffic failed. Now shows **"Handshaking…"** until a handshake completes, then "Connected"; a set `lastError` surfaces as "Blocked" instead of being masked by the running check. The connectivity monitor fires a one-shot `onConnected` callback on the first observed handshake so the status line updates live.
- Verified: idle tunnel → "Handshaking…"; after real traffic → "Connected".

**2. One roam triggered ~3 rebinds.** A single Wi-Fi toggle delivers several capability/connectivity callbacks within a few hundred ms, each spawning its own rebind + DNS re-resolution. Coalesced with an in-flight guard + 1s debounce (the single rebind reads the latest network state when it runs).
- Verified: Wi-Fi off now logs 1 `rebinding WG sockets` + 2 `coalesced` (was 3 rebinds); traffic still flows after the roam.

**3. `Dropping a GlobalRef in a detached thread` on teardown.** The tokio worker threads weren't attached to the JVM, so the Java `GlobalRef`s held by the callbacks warned on drop. Attach each worker via `on_thread_start` (`tunnel.rs` stays JNI-free; the JNI layer supplies the attach closure, matching the existing `with_env` pattern).
- Verified: the same `gotatun device stopped` teardown that logged 2 warnings before now logs 0.

Build clean (`assembleGithubDebug`, all four ABIs; `cargo check`). Existing `WgConnectivityChecker` tests unaffected — the new monitor param is defaulted and the checker itself is untouched.

### Left alone (out of scope, not regressions)
- **Fail-closed copy vs. LAN routing:** RFC1918 destinations bypass the VPN by route, so "no fall-through to direct" is slightly generous for LAN traffic — but that's standard NetGuard/Android routing, not introduced here.
- **DNS TTL clamp** (`259200000`): matches the legacy path exactly; changing it would alter pre-existing behaviour.

### Not covered by this run (emulator limits)
Multi-hour uptime, battery, real Wi-Fi↔cellular roaming with distinct egress IPs, and IPv6 endpoints — the PR's remaining "not yet done" items still stand for these.

PR #595, opened 2026-07-04
[codex] Add on-device APK patcher

kasnder:
## Summary

Adds an on-device APK patcher for traffic inspection builds. The patcher wires a per-app details action through a flavor-gated patching engine that can merge split APKs, patch common certificate pinning implementations, inject a user-CA-trusting network security config, re-sign the result, and hand it to Android's installer.

Also includes supporting documentation and WireGuard flow/cache adjustments that keep inspection-related traffic handling efficient while the WireGuard hijack path is active.

## User Impact

Users of fdroid/github builds get a new "Patch for traffic inspection" action in app details. Play builds keep a no-op fallback and do not include the patcher dependencies. The UI now warns that patched APKs are re-signed and may require uninstalling the original app before installation.

## Validation

- `./gradlew :app:compileFdroidDebugJavaWithJavac :app:compilePlayDebugJavaWithJavac`
- `git diff --check`

I also started `./gradlew :app:assembleFdroidDebug`, but stopped it after it sat silent during packaging; no failure was reported before interruption.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -166,6 +166,11 @@ android {
             // libwgbridge.so built from wgbridge-rs/ by the wgbridgeBuild task
             jniLibs.srcDirs += "$buildDir/rustJniLibs"
         }
+        // The APK-patcher engine (apksig + ARSCLib) is only compiled into the
+        // fdroid and github flavors. The play flavor excludes it and falls
+        // back to NoOpPatcher at runtime via PatcherFactory.
+        fdroid.java.srcDirs += 'src/patcher/java'
+        github.java.srcDirs += 'src/patcher/java'
     }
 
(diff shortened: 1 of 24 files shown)


kasnder:
Closing this draft. The on-device APK patcher is a heavyweight, expert-only capability (merge splits, defeat cert pinning, inject a user-CA network-security-config, re-sign, reinstall) that runs against the app's core simplicity/rigour ethos and its 'no SSL interception, privacy-preserving by construction' positioning. If we pursue traffic-inspection tooling, it should be a clearly-partitioned, github/fdroid-only, loudly-warned feature — or a separate companion tool — rather than something in the main app surface. Parking it for now.

PR #596, opened 2026-07-07
Batch access/usage DB writes, debounce network reloads, back off WG restarts

kasnder:
## Summary
Implements the four highest-ROI fixes from a battery review of the VPN service, WireGuard egress, and database layer:

- **Batch `updateAccess`/`updateUsage`** (`DatabaseHelper.java`): mirrors the existing `insertLog` batching — repeated writes to the same access row within a 2s/50-entry window coalesce into one transaction (last-write-wins for state, summed deltas for usage), and `updateUsage`'s old SELECT-then-UPDATE collapses into a single `UPDATE ... SET sent = COALESCE(sent,0) + ?`. Access-table readers/writers flush pending batches first for read-your-writes consistency; `onDestroy` flushes on shutdown so nothing is lost.
  - Scoped down from the original review: `insertDns` is intentionally left unbatched, since its return value synchronously gates `prepareUidIPFilters` (firewall rule enforcement for newly resolved trackers) — batching it would introduce a correctness-relevant delay.
- **Serialize + back off WireGuard recovery** (`WgEgress.kt`): network-change rebinds now run through a single-thread executor with dedup instead of spawning an unbounded raw `Thread` per event. The broken→full-restart loop now backs off exponentially (20s → 40s → 80s → ... capped at 5 min) instead of retrying every ~20s forever on a captive portal or blocked UDP path. Backoff resets on any confirmed fresh handshake.
- **Debounce network reloads** (`ServiceSinkhole.java`): bursts of `ConnectivityManager` callbacks (common during Wi-Fi↔cellular handoffs) now coalesce into a single reload 1.5s after the burst settles, instead of restarting the VPN/WireGuard per callback. Cancelled cleanly on service destroy.
- **Bounded wakelock**: `onStartCommand`'s wakelock now has a 30s timeout as a safety net against a hung or throwing command path holding it indefinitely.

## Test plan
- [x] `./gradlew compileFdroidDebugJavaWithJavac` and `compileFdroidDebugKotlin` — clean
- [x] `./gradlew testFdroidDebugUnitTest` — full suite passes
- [x] Added `DatabaseHelperAccessBatchTest` covering coalescing (last-write-wins), usage accumulation, and distinct-key isolation within a batch
- [ ] Not yet exercised on a real device — recommend a manual pass (network flapping, WireGuard recovery on a blocked/captive network, and traffic-heavy app usage) before merging, especially given the timing-sensitive WireGuard recovery changes

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/DatabaseHelper.java
+++ b/app/src/main/java/eu/faircode/netguard/DatabaseHelper.java
@@ -39,6 +39,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -50,6 +51,8 @@ public class DatabaseHelper extends SQLiteOpenHelper {
     public static final int ACCESS_UNCERTAIN_MULTIPLE_TRACKERS = 3;
 
     public Cursor getHosts(int uid) {
+        flushAccessBatch();
+        flushUsageBatch();
         lock.readLock().lock();
         try {
             SQLiteDatabase db = this.getReadableDatabase();
@@ -63,6 +66,8 @@ public class DatabaseHelper extends SQLiteOpenHelper {
     }
 
     public Cursor getHosts() {
+        flushAccessBatch();
+        flushUsageBatch();
         lock.readLock().lock();
         try {
             SQLiteDatabase db = this.getReadableDatabase();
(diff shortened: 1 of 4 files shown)


kasnder:
For context: this PR only covers fixes 1-4 from the battery review. Not addressed here, in case they're worth follow-up PRs:

- **`blockKnownTracker` read-lock scope** — `dh.getQAName(uid, daddr, true)` still runs a DB cursor loop while holding the service's read lock on every tracker-IP cache miss, blocking writers during that query.
- **`System.gc()` after every command** (`ServiceSinkhole.java`) — forces a stop-the-world GC on every start/reload/watchdog/householding; safe to remove and let ART schedule collection itself.
- **Missing index on `access.time`** — `getInsightsData7Days`/`getRecentTrackerActivity` filter on `time` with no supporting index, so they run full/partial table scans; the table also has no periodic pruning (only `log` and `dns` get cleaned up). Landing an index here trades a small write-side cost (each access upsert) for query wins on the Insights/Timeline screens.
- **`WgConnectivityMonitor`'s fixed 1s poll interval** — not urgent (no wakelock is held, so it doesn't block doze), but could go adaptive (e.g. 5-15s when healthy/idle) for a modest win.

Also worth noting, since it didn't fit clean batching: `insertDns` has the same per-call transaction overhead as `updateAccess`/`updateUsage` did, but its return value synchronously drives `prepareUidIPFilters` (rebuilding firewall IP filters), so batching it risks delaying enforcement for freshly resolved trackers. Left as-is; would need a design that preserves the synchronous "is this new" signal while deferring only the persisted write.

kasnder:
## Code-review fixes + emulator smoke test (through a483b59)

Applied the findings from a review of this branch, plus one deferred battery item, and ran an emulator pass.

### Fixes pushed

Five review findings, all in the new batching / WG-recovery mechanics:

- **Rebind events are now coalesced, not dropped.** The old `rebindQueued` skip meant a second network change during an in-flight rebind was discarded, potentially leaving sockets bound to a network that had already gone away. Replaced with in-flight + dirty flags so the executor re-runs once on the current network.
- **Restart backoff resets on a confirmed-fresh handshake even under generation churn.** A concurrent `onUnderlyingNetworkChanged` (which bumps `verificationGeneration`) could suppress the `restartAttempts = 0` reset, leaving the *next* transient breakage stuck behind a stale backoff of up to 5 min. The reset now happens ahead of the generation guard, gated only on a genuinely fresh handshake.
- **The delayed (backed-off) restart no longer fires redundantly.** `pendingRestartRunnable` now checks `forceRestartPending` first, so a restart already handled by another recovery path (e.g. a network-change reload) does not kick a needless restart of a healthy tunnel.
- **Access/usage batches get a bounded time-based flush.** Previously the tail of a burst stayed invisible (no `notifyAccessChanged`) until the next reader or shutdown, and was lost outright on an unclean kill. The first entry of an idle batch now schedules a flush on the DB handler thread, bounding both staleness and the loss window to ~2s.
- **A specified `block` value is preserved across last-write-wins coalescing** when a later same-key update omits it (latent today, since production always passes `block = -1`). Added a regression test.

Plus the deferred item from the battery review:

- **Removed the forced `System.gc()` after every command** (start/reload/stop/watchdog/householding). Let ART schedule collection itself; the explicit call only added latency/jank on the command path.

**Build/unit verification:** Java + Kotlin compile clean and the full `testFdroidDebugUnitTest` suite passes (including the new block-preservation test).

### Emulator smoke test (Small_Phone AVD, Android 16)

Built and installed the full `fdroidDebug` APK (Rust `wgbridge` native lib for all 4 ABIs included) and exercised it live:

- ✅ App installs & launches, `libnetguard.so` loads, VPN service starts and establishes `tun0` (both WG mode MTU 1420 and plain-filter MTU 10000 observed).
- ✅ Fail-closed WG init did **not** block startup — service ran, "WireGuard — Connected" in settings.
- ✅ **Live capture → DB write pipeline works** under the changed code: 70 `dns` rows + 263 `app` rows written by the running service; resolved tracker domains (`doubleclick.net`, `graph.facebook.com`, `app-measurement.com`, `region1.google-analytics.com`) landed in `dns`.
- ✅ `WgConnectivityMonitor` is live and handled a real **cellular→WiFi** transition without crashing.
- ✅ Service ran through many command/reload cycles (exercising the `System.gc()` removal + reload debounce).
- ✅ **Full-session crash/ANR/native-fault/ACRA scan came back empty — zero crashes.**

**Honest limitations of the emulator pass:**

- The batched `updateAccess` path did not write live `access` rows: that call is gated on `isTracker` **and** a data packet whose IP reverse-resolves to a tracker hostname, and `updateUsage` needs `track_usage` (off by default). `am start`-driven Chrome doesn't reliably complete that. The gating lives in unchanged service code; the batching/coalescing/flush/block-preservation logic itself is covered deterministically by the unit test.
- **The WG recovery changes (the first three fixes) were the least exercised** — WG egress was only briefly the active path; most of the session ran plain NetGuard filtering, where the rebind/backoff code is dormant. The emulator does **not** substitute for the on-device WG-recovery pass.

### Still outstanding before merge

- **Manual on-device pass is still required and remains the real validation for the WG changes** — network-flapping / captive-portal / blocked-UDP scenarios against this updated branch.
- **The other follow-up items remain out of scope** (documented in the PR discussion): `blockKnownTracker` read-lock scope, the missing `access.time` index plus access-table pruning, `WgConnectivityMonitor`'s fixed 1s poll, and the un-batched `insertDns` transaction overhead.

PR #599, opened 2026-07-11
Strip build-time metadata from native libs for reproducible builds

### LINKED ISSUE #440: Question on permissions
Nice to see the update, Konrad! While it rolled in, it triggered a report at our scanners at IzzyOnDroid, maybe you can help us with clarification here so we can put the points on the app's "green list" with the proper explanation:

```
! repo/net.kollnig.missioncontrol_2024121502.apk declares intent-filter(s): android.net.VpnService
! repo/net.kollnig.missioncontrol_2024121502.apk declares sensitive permission(s):
  android.permission.READ_PHONE_STATE android.permission.QUERY_ALL_PACKAGES
! repo/net.kollnig.missioncontrol_2024121502.apk contains signature block blobs: 0x504b4453 (DEPENDENCY_INFO_BLOCK; GOOGLE)
```

I've just added

```yaml
android.net.VpnService: needed to filter tracker access
```

as that's how the app works. I have no idea what it needs `READ_PHONE_STATE` for, though. `QUERY_ALL_PACKAGES` seems clear again: to choose apps to be filtered. So I've put:

```yaml
android.permission.QUERY_ALL_PACKAGES: needed to list apps to filter trackers for
```

Last but not least, there's `DEPENDENCY_INFO_BLOCK`, which can easily be avoided by a minor addition to your `build.gradle`:

```gradle
android {
    dependenciesInfo {
        // Disables dependency metadata when building APKs.
        includeInApk = false
        // Disables dependency metadata when building Android App Bundles.
        includeInBundle = false
    }
}
```

For some background: that BLOB is supposed to be just a binary representation of your app's dependency tree. But as it's encrypted with a public key belonging to Google, only Google can read it – and nobody else can even verify what it really contains. More details can be found e.g. here: [Ramping up security: additional APK checks are in place with the IzzyOnDroid repo](https://android.izzysoft.de/articles/named/iod-scan-apkchecks#blobs).

So far:

image

Would be great if you could help us fill the gaps (especially for "chocolate warnings"), and get that blob removed.

PS: I've just checked my notes why TrackerControl is not enrolled for our RBs at IzzyOnDroid. It seems we were able to build the app, but there were differences in the `*.so` files keeping it from being RB. If you're interested to have the app enrolled (no signing change here as we already ship your builds), we gladly give it another try – for which it would help to know how your setup looks like, which tools and versions you are using and, if it's no secret, which path you build from (i.e. "the build directory location"). If you prefer, I'd open a separate issue for that then.

kasnder:
## Summary
- Strips the `.comment` section (embeds clang's version string, including the "+bolt"/"-bolt" markers that kept showing up in @IzzySoft's diffoscope diffs) and `.note.gnu.build-id` from `libnetguard.so` post-link, per [IzzyOnDroid's RB debugging guide](https://izzyondroid.org/docs/reproducibleBuilds/DebugFailedRBs/#clang).
- Applies the same treatment to `libwgbridge.so` (the Rust WireGuard bridge, added after this thread started), plus `--remap-path-prefix` for the cargo registry/checkout paths and `-Wl,--build-id=none`, mirroring the existing `-ffile-prefix-map` fix for the C sources.

## Verification
Built the app twice from a clean state and diffed the resulting APKs:
```
differing entries: 0
```
Also confirmed with `llvm-objdump`/`strings` on the packaged `.so`s in both APKs: no `.comment`, no `.note.gnu.build-id`, no leaked absolute build paths.

Closes #440 — would appreciate @IzzySoft re-testing RB against this branch when convenient.

## Test plan
- [x] `./gradlew assembleGithubRelease` succeeds
- [x] Two independent clean builds produce byte-identical APKs (0 CRC differences across all zip entries)
- [x] `llvm-objdump -h` confirms `.comment`/`.note.gnu.build-id` absent from both native libs in the packaged APK

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -32,4 +32,15 @@ target_link_libraries( netguard
 target_link_options(netguard PRIVATE
     "-Wl,-z,max-page-size=16384"
     "-Wl,-z,common-page-size=16384"
-    "-Wl,--build-id=sha1")
+    "-Wl,--build-id=none")
+
+# The clang version string embedded in .comment (which varies with the NDK's
+# use of BOLT-optimized toolchains, "+bolt"/"-bolt") and any residual
+# .note.gnu.build-id both harm reproducibility. Strip them post-link.
+# See https://izzyondroid.org/docs/reproducibleBuilds/DebugFailedRBs/#clang
+add_custom_command(TARGET netguard
+    POST_BUILD
+    COMMAND ${CMAKE_OBJCOPY} --remove-section .comment
+            --remove-section .note.gnu.build-id
+            $<TARGET_FILE:netguard>
+    VERBATIM)
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -81,9 +81,16 @@ tasks.register('wgbridgeBuild') {
         env['PATH'] = "${cargoDir}${File.pathSeparator}${System.getProperty('user.home')}/.cargo/bin${File.pathSeparator}${env['PATH'] ?: ''}"
         env['ANDROID_NDK_HOME'] = findNdkDirectory().absolutePath
         // Android 15+ requires 16KB-aligned ELF LOAD segments.
+        // --remap-path-prefix strips the checkout- and machine-specific
+        // absolute paths (cargo registry cache, wgbridge-rs checkout dir)
+        // that rustc otherwise bakes into panic/track_caller strings, so
+        // libwgbridge.so is reproducible across build machines.
         env['RUSTFLAGS'] = ((env['RUSTFLAGS'] ?: '') +
                 ' -C link-arg=-Wl,-z,max-page-size=16384' +
-                ' -C link-arg=-Wl,-z,common-page-size=16384').trim()
+                ' -C link-arg=-Wl,-z,common-page-size=16384' +
+                ' -C link-arg=-Wl,--build-id=none' +
+                " --remap-path-prefix=${System.getProperty('user.home')}/.cargo=/cargo" +
+                " --remap-path-prefix=${wgbridgeSrcDir.absolutePath}=/wgbridge-rs").trim()
 
         // cargo-ndk drives the per-ABI cross-compilation and linker setup.
         def cargoNdk = "${System.getProperty('user.home')}/.cargo/bin/cargo-ndk"
@@ -111,6 +118,28 @@ tasks.register('wgbridgeBuild') {
         // gotatun statically; drop the stray artifacts.
         wgbridgeAbis.each { abi ->
             fileTree(dir: "$wgbridgeOutDir/$abi", include: 'libgotatun*.so').each { it.delete() }
+        }
+
+        // The NDK's clang linker still embeds a .comment section (compiler
+        // version string, incl. "+bolt"/"-bolt" markers that vary by NDK
+        // build) even with --build-id=none. Strip it for reproducibility,
+        // matching the equivalent objcopy step for the CMake-built netguard
+        // library. See https://izzyondroid.org/docs/reproducibleBuilds/DebugFailedRBs/#clang
+        def hostTag = { ->
+            def os = System.getProperty('os.name').toLowerCase()
+            if (os.contains('mac')) return 'darwin-x86_64'
+            if (os.contains('win')) return 'windows-x86_64'
+            return 'linux-x86_64'
+        }()
+        def objcopy = "${findNdkDirectory().absolutePath}/toolchains/llvm/prebuilt/${hostTag}/bin/llvm-objcopy"
+        wgbridgeAbis.each { abi ->
+            def lib = file("$wgbridgeOutDir/$abi/libwgbridge.so")
+            if (lib.exists()) {
+                providers.exec {
+                    commandLine objcopy, '--remove-section', '.comment',
+                            '--remove-section', '.note.gnu.build-id', lib.absolutePath
+                }.result.get().assertNormalExitValue()
+            }
         }
     }
 }


hiSandog:
Removing build IDs and remapping source paths directly addresses the two common sources of native-library differences described here. One additional check could build from two different absolute checkout paths, not only two clean states, because path leakage often remains hidden when both builds use the same workspace location. It may also be worth recording the minimum Android NDK and Rust toolchain versions that support the chosen strip and linker flags so older contributor environments fail with a clear compatibility explanation.

kasnder:
Thanks, @hiSandog Been struggling with this for a long time. Will look into it..

kasnder (commented):
**Verdict: recommend closing, not merging — a hardened version of this exact change already landed on `master` as `3562aa3` ("Fixes #440", and #440 is now closed), this PR is `CONFLICTING`, and its `RUSTFLAGS` handling is a regression relative to what shipped.**

The technique itself is sound (strip `.comment` + drop `.note.gnu.build-id`/build-id, remap Cargo paths), and the CMake half of this diff is byte-identical to what's already on `master`. The problem is specifically the Rust `RUSTFLAGS` block, which the maintainer already rewrote in `3562aa3` to use `CARGO_ENCODED_RUSTFLAGS`.

### Blocking (for merging as-is)
- **Superseded / redundant.** `origin/master` already contains `3562aa3` "Strip build-time metadata from native libs for reproducible builds" (Fixes #440); issue #440 is closed. `git diff 3562aa3 <pr-head>` on the two changed files shows the *only* delta is the `RUSTFLAGS` block in `app/build.gradle` — and the merged form is the better one. Merging this PR would revert that hardening. Suggest closing the PR (or re-targeting it to something not already covered).
- `app/build.gradle:84` — setting `env['RUSTFLAGS']` is silently ignored by Cargo whenever `CARGO_ENCODED_RUSTFLAGS` is already present in the environment (Cargo consults the encoded form first and never falls back to `RUSTFLAGS` when it's set). If that variable is set by the surrounding env — or injected by the `cargo ndk` invocation on line ~99-107 — the remap/build-id flags are dropped and reproducibility silently breaks with no error. The merged `3562aa3` avoids this exactly by reading either `CARGO_ENCODED_RUSTFLAGS` or `RUSTFLAGS`, appending, and re-emitting the encoded form. *(The cargo-ndk-sets-it path is unverified for this cargo-ndk version, but the merged code's explicit handling of a pre-existing `CARGO_ENCODED_RUSTFLAGS` shows the maintainer treated it as real.)*

### Should fix (were this re-applied instead of the merged version)
- `app/build.gradle:88-89` — `RUSTFLAGS` is whitespace-tokenized by Cargo, so `--remap-path-prefix=${user.home}/.cargo=/cargo` breaks if `$HOME` or the checkout dir contains a space (the flag splits into malformed tokens). The merged version's ``-encoded form is space-safe.
- `app/build.gradle:88` — hardcodes `${user.home}/.cargo` and ignores `CARGO_HOME`; a custom `CARGO_HOME` leaves registry paths unremapped, leaking machine-specific paths. The merged version falls back to `~/.cargo` only when `CARGO_HOME` is unset.
- Cross-machine reproducibility gap (present in the merged version too, so a follow-up rather than this PR's fault): the NDK is pinned (`wgbridgeNdkVersion = '27.2.12479018'`), but there is no `rust-toolchain.toml`. The commit message claims "reproducible across build machines," yet a rebuilder on a different `rustc` gets different codegen and `/rustc/<hash>/` std paths. Consider pinning the Rust toolchain to fully close #440 for `libwgbridge.so`.

### Nits
- `app/CMakeLists.txt:43` (and the Rust `objcopy` loop) — `--remove-section .note.gnu.build-id` is redundant now that `--build-id=none` is set (`CMakeLists.txt:35`); the section won't be emitted. Harmless/defensive, and the comment already says "residual."
- `CMakeLists.txt:35` `--build-id=none` and the `POST_BUILD` `objcopy` apply to *all* configs, including the `netguard` debug build, so native crash tombstones lose build-id correlation there too. Standard RB tradeoff and DWARF debug info is untouched, so this is fine — just noting it isn't scoped to release.
- Verification methodology: two clean builds on the *same* machine (same `$HOME`, same checkout path) produce identical embedded paths regardless of whether the remap works, so the "0 differing entries" evidence doesn't actually exercise `--remap-path-prefix`. The real test is a rebuild under a different home/checkout path (what IzzyOnDroid does) — which is precisely where the `RUSTFLAGS`-vs-encoded difference matters.

### What's good
- Correctly identifies the two real RB blockers — the clang version string in `.comment` (incl. `+bolt`/`-bolt`) and the build-id — matching IzzyOnDroid's clang guidance.
- The `.comment` strip is genuinely necessary even though `wgbridge-rs/Cargo.toml` sets `[profile.release] strip = true`: symbol stripping does not remove `.comment` (`-R`/`--remove-section` is required), so the extra `objcopy` pass is doing real work.
- The `hostTag` helper correctly hardcodes `*-x86_64` for every host — NDK r27 ships only x86_64 host binaries (`darwin-x86_64` runs under Rosetta on Apple Silicon; `linux-x86_64` on arm64 Linux), so this is the right resolution, not a bug.

kasnder:
Review update (2026-07-11): **close this PR without merging.** The intended reproducible-build fix already landed on `master` in `3562aa3`, and this branch is now conflicting. The merged implementation is also safer: it preserves/encodes existing Cargo flags via `CARGO_ENCODED_RUSTFLAGS`, handles spaces in paths, and respects `CARGO_HOME`; merging this branch would regress that hardening. No further code changes are needed here.

kasnder:
Closing as superseded: the safer implementation already landed on `master` in `3562aa3`, including robust `CARGO_ENCODED_RUSTFLAGS` handling.

Issue #420, opened 2024-05-08
Instagram not visible in list of apps. 


soliverc:
**Do you use TrackerControl from Google Play/F-Droid/other?**
F-Droid

**Describe the bug**
Instagram is not visible in the app list, but it is being blocked. I cannot receive direct messages from people unless I disable Tracker Control entirely.

**To Reproduce**
1. Install tracker control
2. Try to have a conversation with somebody on instagram direct message
3. Not possible - messages are blocked
4. Check tracker control to adjust settings for Instagram
5. Instagram is not in the list of apps
6. The only way to fix this problem is to disable tracker control for your entire phone

kasnder:
Does this issue still exist in the latest app? @soliverc

soliverc:
Hi, I don't use Instagram anymore so can't check.


________________________________

Issue #229, opened 2021-07-18
Android groups applications on UID


Halilb1978:
System applications  have the same UID, even they have a different package and application name. Changing a setting in an app effects also other apps in same UID. For example disabling internet access of an app also disables the others. It would be nice to list/see/filter apps with UID option.

kasnder:
Yeah, that's right. The Android VPN doesn't differentiate between apps with the same UID. What exactly do you want to do differently? Can you explain some more?

Halilb1978:
Just a listing option to see which apps are in the same UID.

AnanasPfirsichSaft:
NetGuard can do this. And it is really useful. Perhaps it can be done as a search filter like "uid:12345? This wouldn´t clutter the UI too much and of course the UID has to be shown in the app details screen.

I appreciate the simplicity of TC. Don´t add to much... ;)

kasnder:
This is quite an expert feature. Won't implement.

PR #611, opened 2026-07-11
Make filtered-but-hidden apps reachable via search (#420, #483)

kasnder:
## Problem

Some installed apps are **filtered/blocked by TC but never appear in TC's app list**, so the user cannot open them to add a per-tracker exception or exclude the app. The only workaround is disabling TC entirely — the worst kind of (unescapable) breakage.

- **#420** — Instagram is blocked (DMs won't send) but is missing from the list; searching "Instagram" finds nothing.
- **#483** — Feeder (RSS reader, F-Droid) is installed but absent from the list.
- Related: **#229** "Android groups applications on UID".

## Root cause

Blocking is keyed on **UID** in the packet path (`ServiceSinkhole.isAddressAllowed` / `blockKnownTracker` / `shouldTrackApp`), and VPN routing is decided from `Rule.getRules(true, ...)` (the full app set, gated only by `!apply || (!includeSystem && system)`).

The **visible** list, however, is built by `Rule.getRules(false, ...)`, which applies the `show_system` / `show_frozen` / `show_nointernet` **display filters**. So an app hidden by one of those filters — e.g. an OEM-preinstalled package that TC routes because *"include system apps in VPN"* is enabled, a frozen package, or a shared-UID sibling (#229) — is **still routed and blocked by UID while being absent from the list**.

`AdapterRule`'s search only iterated the already-filtered visible subset (`listAll`), so the hidden-but-filtered app could not be found by search either. That matches #420's exact repro (search "Instagram" → missing).

## Fix (search-only; default view unchanged)

- `Rule.getSearchStubs(context)` — a lightweight, side-effect-free list of rules for **every installed app** (skips tracker-count queries, default seeding, sorting, snackbars). Enough to be matched by search and to open the per-app details screen.
- `AdapterRule` — search now runs over the visible list **plus every other installed app** (pool built lazily, cached, invalidated on `set()`). A hidden-but-filtered app can now be found and opened to configure an exception or exclude it. Visible apps keep their fully-populated rules (tracker counts intact). An **empty query still shows exactly the visible list**, so there is no clutter regression in the default view.

This keeps the recovery path (search → open → configure/exclude) working for any app TC can filter, without adding user-facing toggles (per the project's simplicity-over-configurability philosophy).

## Deferred (documented, not fixed here)

The underlying display-filter behaviour and shared-UID runtime attribution (#229 — the packet path resolves a UID to `getPackagesForUid(uid)[0]`, so per-package settings on a non-representative sibling may not take effect at runtime) are left intact. This change guarantees **reachability/recoverability**, which is the user-facing breakage; a fuller shared-UID attribution redesign is out of scope here.

## Testing

- `./gradlew :app:compileGithubDebugJavaWithJavac` — **BUILD SUCCESSFUL** (Rust `wgbridgeBuild` via preBuild ran; `cargo` present).
- Not verified on-device: I cannot install Instagram/Feeder to reproduce the exact hidden-app state. The change is behaviour-preserving for the default (empty-query) view and only widens the search pool. On-device confirmation that searching a hidden app now surfaces it, and that opening it lets the user exclude/except it, is still recommended.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/AdapterRule.java
+++ b/app/src/main/java/eu/faircode/netguard/AdapterRule.java
@@ -64,6 +64,7 @@ import net.kollnig.missioncontrol.data.InternetBlocklist;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 
 public class AdapterRule extends RecyclerView.Adapter<AdapterRule.ViewHolder> implements Filterable {
@@ -421,7 +422,7 @@ public class AdapterRule extends RecyclerView.Adapter<AdapterRule.ViewHolder> im
             @Override
             protected FilterResults performFiltering(CharSequence query) {
                 List<Rule> listResult = new ArrayList<>();
-                String queryStr = query == null ? "" : query.toString().toLowerCase().trim();
+                String queryStr = query == null ? "" : query.toString().toLowerCase(Locale.ROOT).trim();
                 if (queryStr.isEmpty())
                     listResult.addAll(listAll);
                 else {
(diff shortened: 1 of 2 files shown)


kasnder (commented):
Review finding: the new case-insensitive search path used the device default locale for `toLowerCase()`. In locales with special case mappings (notably Turkish), ASCII package names and user queries can normalize differently, causing hidden apps to remain undiscoverable. I fixed all three normalizations to use `Locale.ROOT` and verified `:app:compileGithubDebugJavaWithJavac`.

PR #615, opened 2026-07-11
Harden WireGuard bridge integration

kasnder:
## Summary

- serialize the native WireGuard socketpair descriptor lifetime across publish, write, detach, and close
- add bounded TCP DNS stream reassembly with sequence, retransmission, gap, timeout, and flow-limit handling
- bind connectivity monitors, recovery work, and delayed reloads to a tunnel generation and object identity
- attach JNI worker threads as daemon threads and keep restart backoff intact across forced restarts

## Root cause

The bridge treated an atomically loaded raw file descriptor as if it owned the descriptor lifetime, parsed DNS-over-TCP only when a full message fit in one segment, and allowed asynchronous recovery results from an obsolete tunnel to affect its replacement.

## Impact

This prevents writes to reused descriptors, restores tracker DNS mapping for segmented TCP responses, and stops stale recovery tasks from restarting a healthy replacement tunnel.

## Validation

- `cargo test --all-targets` (20 tests)
- `cargo clippy --all-targets -- -D warnings`
- `./gradlew :app:wgbridgeBuild :app:externalNativeBuildFdroidDebug :app:testFdroidDebugUnitTest --no-daemon`
- native build completed for arm64-v8a, armeabi-v7a, x86, and x86_64

### DIFF
--- a/app/src/main/java/net/kollnig/missioncontrol/wg/WgEgress.kt
+++ b/app/src/main/java/net/kollnig/missioncontrol/wg/WgEgress.kt
@@ -44,6 +44,13 @@ object WgEgress {
     private const val RESTART_BACKOFF_MAX_MS = 5 * 60_000L
 
     @Volatile private var tunnel: WgTunnel? = null
+    // Monotonically identifies the tunnel instance published in [tunnel].
+    // Recovery runs off-thread and may finish after startOrUpdate has replaced
+    // that instance, so object identity alone is not enough to associate its
+    // result with the lifecycle that launched it. AtomicLong also makes the
+    // read/modify/write safe across the VPN, monitor, and rebind threads.
+    private val tunnelGeneration = java.util.concurrent.atomic.AtomicLong(0)
+    private val tunnelLifecycleLock = Any()
     @Volatile private var lastError: String? = null
     @Volatile private var verificationGeneration: Long = 0
     @Volatile private var currentConfig: String? = null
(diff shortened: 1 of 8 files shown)


kasnder (commented):
Solid, well-reasoned hardening. The C fd-lifetime serialization and the Kotlin generation/identity logic look correct; one should-fix correctness edge remains in the new TCP DNS reassembler.

### Should fix
- `wgbridge-rs/src/dns.rs:114-121` — A TCP segment lying **entirely** before the reassembly frontier (a full retransmission of already-consumed bytes: `already_seen > payload.len()` yet still `< 0x8000_0000`) falls into the forward-gap `else` and **drops the whole flow**, losing reassembly for the rest of that DoT response — exactly the coverage this PR adds. Only a real forward gap (`already_seen >= 0x8000_0000`) should drop the flow; a wholly-old duplicate should skip its bytes and keep the flow. Suggest branching on `already_seen >= 0x8000_0000` for the drop, else `&segment.payload[(already_seen as usize).min(segment.payload.len())..]` (an empty tail is a harmless no-op). Verified by tracing sequence numbers; the `stateful_inspector_ignores_retransmitted_tcp_bytes` test only exercises an *exact* duplicate (`already_seen == payload.len()`), not a wholly-old one, so this slips through.

### Nits
- `app/src/main/jni/netguard/netguard.c:459` (`jni_wireguard_stop`) — on the (practically unreachable) `pthread_mutex_lock` failure it returns without closing `wg_outbound_fd`, leaking the descriptor, whereas `jni_wireguard_start` cleans up its fds on the same failure. Defensible (you can't close without the lock), just inconsistent.
- `write_wireguard_packet` now takes `wg_outbound_lock` on every hijacked packet (previously lockless atomics). The critical section is a single bounded non-blocking `write`, so this is fine — flagging only because egress throughput has been a prior concern.

### What's good
- fd lifetime is now fully serialized under one mutex, and the `ip.c:412-430` reorder is behavior-equivalent to the old `wg_enabled`/`wg_fd` atomics (traced every case) while closing the check-then-write fd-reuse TOCTOU. Fail-closed semantics for the DNS/gap cases are preserved.
- Generation + object-identity snapshotting (`WgEgress.kt:785-804`, `341-356`, `569-584`) cleanly stops stale monitors/recovery tasks from restarting a healthy replacement tunnel and from zeroing `restartAttempts` at `WgEgress.kt:298` — a real battery/backoff-integrity win. `requestReloadCb` is an async `startForegroundService`, so holding `tunnelLifecycleLock` across it can't deadlock against the `monitorLock -> tunnelLifecycleLock` order.
- The reassembler is well-bounded and panic-safe: 64-flow cap with idle eviction, 65537-byte buffer cap, IPv4-fragment/forward-gap rejection, every header index length-guarded, and the Java sink kept under `catch_unwind` (`dns.rs:180-186`). `startOrUpdate` also now publishes `tunnel` under the lock only after the companion `currentConfig`/`currentTunPfd` fields are installed (`WgEgress.kt:255-265`), and closes `rxFd` exactly once in `finally`.

kasnder:
Re-reviewed the latest head, including the fd lifetime serialization, generation/identity gating, and TCP DNS reassembly. The wholly-old retransmission bug from the earlier review is fixed and covered by a focused regression test. The current CI test job passes, and I found no remaining correctness blocker. Recommendation: mark ready and merge this PR first; #618 is stacked on it.

PR #618, opened 2026-07-11
Model Android underlying connectivity

kasnder:
## Stacked on

- #615

This PR should be reviewed and merged after #615; its diff contains only the Android underlying-connectivity model.

## Summary

- combine Android's default-network callback with the `INTERNET + NOT_VPN` callback stream
- take an eager physical-network snapshot and debounce callback bursts for 300 ms
- reduce observations to distinct, process-global connectivity generations
- include default path identity, validation/metering, transports, interface, DNS, routes, and link addresses
- feed generation and connected state into `WgEgress`, discarding stale recovery results and avoiding offline restart churn
- invalidate state around TUN transitions without rebinding an intentionally replaced tunnel
- preserve the existing 1500 ms service reload policy as a separate concern

The design follows Mullvad's Android pattern of combining callback streams, debouncing/distincting state, taking an eager initial snapshot, and invalidating connectivity state around TUN transitions. It intentionally does not add `Network.bindSocket`.

## Impact

Wi-Fi/cellular handoffs, DHCP/DNS changes, and callback bursts now produce one coherent physical-path generation. Secondary candidate changes trigger reevaluation without rebinding when the resolved default path is unchanged. Recovery work is discarded when a newer network generation arrives, and liveness monitoring resumes after an offline path reconnects.

## Validation

- focused `UnderlyingNetworkReducerTest` and `WgEgressRecoveryTest`
- full `./gradlew :app:testFdroidDebugUnitTest`
- `./gradlew :app:assembleFdroidDebug`
- native build completed for arm64-v8a, armeabi-v7a, x86, and x86_64
- `git diff --check`

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -111,10 +111,12 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -159,6 +161,16 @@ public class ServiceSinkhole extends VpnService {
     };
 
     private Object networkCallback = null;
+    private ConnectivityManager.NetworkCallback underlyingDefaultNetworkCallback = null;
+    private volatile Network callbackDefaultNetwork = null;
+    private final java.util.concurrent.atomic.AtomicLong underlyingTunEpoch =
+            new java.util.concurrent.atomic.AtomicLong(0);
+    private static final long UNDERLYING_NETWORK_DEBOUNCE_MS = 300L;
+    private static final Object UNDERLYING_NETWORK_TOKEN = new Object();
+    private final Handler underlyingNetworkHandler = new Handler(Looper.getMainLooper());
+    private final net.kollnig.missioncontrol.wg.UnderlyingNetworkReducer underlyingNetworkReducer =
+            new net.kollnig.missioncontrol.wg.UnderlyingNetworkReducer();
+    private volatile boolean destroying = false;
 
     private boolean registeredInteractiveState = false;
     private PhoneStateListener callStateListener = null;
@@ -1483,6 +1495,7 @@ public class ServiceSinkhole extends VpnService {
     private ParcelFileDescriptor startVPN(Builder builder) throws SecurityException {
         try {
             ParcelFileDescriptor pfd = builder.establish();
+            if (pfd != null) onTunTransition();
             updateUnderlyingNetworks();
             return pfd;
         } catch (SecurityException ex) {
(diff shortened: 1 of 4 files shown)


kasnder (commented):
Solid connectivity model — generation-gated, debounced, and the offline guards genuinely prevent reconnect storms. One correctness issue worth fixing around how the default network is tracked, plus two efficiency nits.

### Should fix
- `ServiceSinkhole.java:3011` and `:3052` — the general `nc` callback is registered via `registerNetworkCallback(INTERNET+VALIDATED)` (`:3093`), so it fires for **every** matching network, not just the system default, yet it assigns `callbackDefaultNetwork = network` for any `NOT_VPN` network. That field is meant to be the *default* path (it's also written authoritatively by `registerDefaultNetworkCallback` at `:3101`/`:3106`) and is consumed as the default-path identity in `sampleUnderlyingNetworks()` (`:3146`). When a non-default validated network emits a callback (Wi-Fi↔cellular handoff, or ROMs/dev-option keeping both radios validated), `callbackDefaultNetwork` flips to it → the reducer sees `defaultChanged` → `rebindRequired=true` → a WG socket rebind + DNS re-resolve on a healthy tunnel. The true default's callback only re-fires on its own changes, so the wrong value can persist, not just flap. Fix: don't set `callbackDefaultNetwork` from `nc`; rely on the dedicated default-network callback (API 24+), and on the `getActiveNetwork()` fallback already in `sampleUnderlyingNetworks()` for pre-N. If you want to keep the `nc` writes, gate them on `network.equals(getActiveNetwork())`.

### Nits
- `ServiceSinkhole.java:3186` — `scheduleUnderlyingNetworkSnapshot` does `reducer.offer(sampleUnderlyingNetworks())` immediately on the (binder) callback thread, but the only consumer (`flush()` at `:3193`) runs inside `emit`, which re-samples at `:3191` first — so the `:3186` value is always overwritten and never contributes to an emitted event. On a flapping network this means one redundant synchronous `getAllNetworks()` + per-network `getNetworkCapabilities`/`getLinkProperties` IPC per callback. Drop line 3186 and let `emit` be the sole sampler (the eager path already re-samples).
- `ServiceSinkhole.java:3172` — `for (int transport = 0; transport <= 8; transport++)` hard-codes the transport range; transports added in future API levels are silently excluded from the signature, so a path change confined to a newer transport type wouldn't register. Minor/forward-compat only.

### What's good
- Process-global monotonic generations (static `NEXT_GENERATION`) correctly survive `ServiceSinkhole` recreation, so stale recovery is discarded; the `!underlyingNetworkConnected` guard in `onMonitorBroken` (WgEgress.kt:318) plus the mid-flight generation/offline re-check in the rebind loop (WgEgress.kt:543) genuinely avoid churning sockets against an absent network.
- Reducer distinctness + 300 ms debounce collapse callback bursts to one coherent generation, and `rebindRequired=false` for secondary-candidate changes avoids rebinding a healthy tunnel — well covered by `UnderlyingNetworkReducerTest`.
- Clean lifecycle: `destroying` gates new scheduling, all three callbacks + handler messages are torn down in `onDestroy` with try/catch, the eager onCreate snapshot seeds `underlyingNetworkConnected` before the tunnel starts, and `startMonitor()` under `rebindLock` is safe since `WgConnectivityMonitor.stop()` interrupts without joining.

kasnder:
Re-reviewed the latest head and the stack relationship. The default-network callback ownership issue and redundant snapshot are fixed; the reducer/recovery-generation behavior looks sound, and I found no remaining code-level blocker. This head is still based on #615’s earlier commit `b95b44f` and does not include #615’s latest fix `9bb5f6c`, and this head currently has no CI check run. Recommendation: merge #615 first, then rebase this branch onto the merged/latest #615 result, retarget to `master`, run CI, and merge once green.

PR #619, opened 2026-07-11
Add bounded VPN connectivity recovery

kasnder:
## What changed

- Add event-driven VPN connectivity checks after startup and physical network transitions.
- Probe the VPN path first, then use a protected bypass socket only when that path fails so device connectivity can be distinguished from a stuck VPN data path.
- Require three consecutive VPN-only failures before repair and allow at most two repairs per network epoch.
- Skip checks while the device is inactive or offline, and cancel pending work on stop, native fatal exit, and service teardown.
- Add focused unit coverage for failure streaks, offline results, budget exhaustion, and network-epoch resets.

## Why

The service already reloads around known network changes, but it cannot detect a data path that remains unusable while the underlying device connection still works. This adds bounded recovery without a permanent polling loop and without competing with native fatal-exit handling.

## Impact

Users can recover automatically from a wedged VPN path after startup or a network handoff. Ordinary device-offline states do not trigger VPN restarts, and all probing is limited to short event-driven sequences while the device is interactive.

## Validation

- `./gradlew testGithubDebugUnitTest`
- `./gradlew assembleGithubDebug`
- `git diff --check`

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -2508,6 +2508,9 @@ public class ServiceSinkhole extends VpnService {
                                         statsHandler.sendEmptyMessage(interactive ? MSG_STATS_START : MSG_STATS_STOP);
                                     }
                                 });
+
+                        if (last_interactive)
+                            reArmVpnHealthCheckIfDeferred();
                     } catch (Throwable ex) {
                         Log.e(TAG, ex.toString() + "\n" + Log.getStackTraceString(ex));
 
@@ -3080,6 +3083,14 @@ public class ServiceSinkhole extends VpnService {
     private final VpnHealthRecoveryPolicy vpnHealthPolicy = new VpnHealthRecoveryPolicy();
     private int vpnHealthGeneration;
     private boolean vpnHealthProbeInFlight;
+    // Set when a probe is skipped purely because the screen was off. A brief
+    // screen-off during the scheduled delay would otherwise silently drop the
+    // whole recovery sequence until the next epoch (e.g. next real network
+    // change), leaving a genuinely wedged VPN path unrepaired for a long
+    // time. Re-armed defensively on the next screen-on, without touching the
+    // epoch/trigger model: this only ever schedules the same probe that was
+    // skipped, at the same generation-check guarded entry point.
+    private boolean vpnHealthProbeDeferredForScreenOff;
 
     private void reloadAfterNetworkChange(final String reason) {
         networkReloadDebounceHandler.removeCallbacksAndMessages(NETWORK_RELOAD_TOKEN);
@@ -3128,8 +3139,16 @@ public class ServiceSinkhole extends VpnService {
                 }, VPN_HEALTH_TOKEN, SystemClock.uptimeMillis() + 1_000L);
                 return;
             }
-            if (!Util.isInteractive(this) || !Util.isConnected(this)) {
-                Log.i(TAG, "VPN health check skipped: device inactive or offline");
+            if (!Util.isInteractive(this)) {
+                // Don't drop the sequence: re-check as soon as the screen
+                // turns back on, still gated on the same generation/state
+                // checks above (and on isConnected below).
+                Log.i(TAG, "VPN health check deferred: screen off");
+                vpnHealthProbeDeferredForScreenOff = true;
+                return;
+            }
+            if (!Util.isConnected(this)) {
+                Log.i(TAG, "VPN health check skipped: device offline");
                 return;
             }
             vpnHealthProbeInFlight = true;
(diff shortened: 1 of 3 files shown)


kasnder (commented):
Solid, well-factored addition: the pure policy is clean and genuinely bounded *per epoch*, offline is correctly distinguished from a wedged path, and lifecycle teardown is handled. The bound has one seam worth pinning down before merge — the recovery budget is reset by the same reload the recovery itself issues — plus an uncoordinated overlap with WgEgress's existing backoff. Neither is fully verifiable from source alone (flagged below).

### Should fix
- **The per-epoch bound can be re-armed by recovery's own reload (the crux).** `VpnHealthRecoveryPolicy.beginEpoch()` (`VpnHealthRecoveryPolicy.java:22`) zeroes `recoveries`, and it is called from `scheduleVpnHealthCheck(true, …)` — invoked on startup and on **every** `reloadAfterNetworkChange` (`ServiceSinkhole.java`, added ~line 3092). The `RECOVER` branch (`finishVpnHealthProbe`, added ~3160) issues a full `reload("VPN connectivity recovery", …)` that re-establishes the tun. The network callback `nc` that drives `reloadAfterNetworkChange` (`ServiceSinkhole.java:2926`) matches `INTERNET`+`VALIDATED` with no `NOT_VPN` and is filtered only by `isActiveNetwork()` (`:3089-3091`, = `getActiveNetwork()`, which is the VPN itself when up). So *if* re-establishing the tun re-fires that validated/active callback, it calls `beginEpoch()` → budget back to 0 → `MAX_RECOVERIES_PER_EPOCH=2` is defeated and a persistently-wedged path can storm-restart. This is the same class as the documented "counter reset on the wrong signal" backoff-defeat. **(unverified** — pre-existing `reload()` does not appear to self-trigger `nc` infinitely, so in practice this may rarely fire; please confirm on-device with logs.) If confirmed it is blocking. Fix: gate `beginEpoch()` on a real change of the *underlying, non-VPN* network identity, or add a wall-clock cap on recoveries independent of epoch.
- **Uncoordinated with WgEgress's own recovery/backoff.** A health-triggered full `reload` rebuilds WireGuard; the fresh-tunnel/teardown path runs `clearRecoveryState()`, which sets `restartAttempts = 0` (`WgEgress.kt:643`) — the very backoff the Jul-8 fix deliberately keyed off *observed rx* rather than handshake. So an external health reload resets Wg's restart backoff, and two recovery mechanisms (WgConnectivityMonitor at ~1s cadence, `WgConnectivityMonitor.kt:226`; vs. this check at 3×5s) can both act on one wedged path. **(unverified** exact reset chain.) Fix: skip the health check while WgEgress has a restart pending (or route both through one owner).
- **Silent abort on a momentary screen-off/offline.** `startVpnHealthProbe` returns without rescheduling when `!Util.isInteractive || !Util.isConnected` (added, `startVpnHealthProbe`). A brief screen-off during the 4–5 s scheduled delay drops the *entire* recovery sequence until the next epoch, so a genuinely wedged path can go unrepaired. Consider re-arming on the next screen-on.

### Nits
- `scheduleVpnHealthCheck` removes only the delayed `VPN_HEALTH_TOKEN`, not the outer immediate `vpnHealthHandler.post(...)` wrappers; under Wi-Fi/cell flapping several queue up, each starting a new epoch and resetting the budget. Harmless on its own but compounds the first item.
- `canReachHealthEndpoint` probes hardcoded `9.9.9.9`/`1.1.1.1:443` (`VPN_HEALTH_ENDPOINTS`). If a user's WireGuard/DNS config blackholes those two IPs, the VPN probe fails even when the path is healthy → false recovery. Prefer the already-configured `validate` host (`ServiceSinkhole.java:2615`) or the WG endpoint.
- The risky wiring (the `beginEpoch()` call sites) has no test coverage; the pure policy is well tested, and `newNetworkEpochRestoresRecoveryBudget` actually *encodes* the reset-on-epoch behavior as desired — which is exactly the double-edged property above.

### What's good
- `VpnHealthRecoveryPolicy` is a pure, `synchronized`, well-unit-tested state machine (streaks, offline, exhaustion, epoch reset all covered).
- Distinguishes a wedged VPN data path from a genuinely offline device via a *protected bypass* socket, so ordinary offline states don't trigger restarts.
- Battery-conscious and lifecycle-safe: event-driven (no polling loop), gated on screen-on/connected, cancelled on `stop`/`nativeExit`/`onDestroy`, with a generation counter + in-flight guard against overlapping probes.

kasnder:
Blocking: the “VPN” probe does not traverse the VPN. TrackerControl explicitly excludes its own package from the `VpnService` builder (`addDisallowedApplication(getPackageName())`), so a plain `Socket` opened by `ServiceSinkhole` already bypasses the VPN. Consequently `canReachHealthEndpoint(false)` and the protected fallback both test the underlying network; when the device is online the first succeeds even if the TUN/native/WireGuard data path is wedged, and the policy records the VPN as healthy. Please redesign around a signal that actually traverses or observes the tunnel (for example native/WireGuard data-plane telemetry or a demonstrably VPN-bound probe) and add an integration/on-device test proving the two paths diverge. Until then this feature cannot achieve its stated purpose; rework substantially or close rather than merging.

kasnder:
Closing rather than applying a cosmetic fix: because TrackerControl excludes its own package from the VPN, both Java socket probes use the underlying network and cannot distinguish a healthy tunnel from a wedged one. A future replacement needs a genuinely tunnel-observing/native data-plane signal and an integration test proving path separation.

PR #621, opened 2026-07-11
Make VPN replacement fail closed

kasnder:
## Summary

- activate a temporary catch-all blocking tunnel before stopping the current packet engine
- establish and validate the configured tunnel before closing the blocking descriptor
- keep traffic fail-closed when either establishment stage fails
- add focused tests for successful sequencing and each failure boundary

## Why

The non-handover replacement path stopped the native worker and closed the active VPN descriptor before creating its replacement. During that interval Android could route application traffic directly over the underlying network.

The new sequence always leaves an active VPN interface in place. The packet engine and encrypted egress bridge are rebound only after the configured replacement descriptor is ready.

## Impact

VPN configuration changes can briefly pause routed traffic, but they no longer introduce a direct-routing window. Excluded applications retain their existing bypass behavior during replacement.

## Validation

- `./gradlew testFdroidDebugUnitTest --tests eu.faircode.netguard.VpnReplacementSequencerTest --no-daemon`
- `./gradlew testFdroidDebugUnitTest --no-daemon`

Both commands passed on the implementation snapshot. A full rerun on the latest base is rebuilding the native libraries and remains in progress at publication time.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -631,41 +631,29 @@ public class ServiceSinkhole extends VpnService {
                 } else {
                     last_builder = builder;
 
-                    boolean handover = prefs.getBoolean("handover", false);
-                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
-                        handover = false;
-                    Log.i(TAG, "VPN restart handover=" + handover);
-
-                    if (handover) {
-                        // Attempt seamless handover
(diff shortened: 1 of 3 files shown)


kasnder:
No blocking findings. The fail-closed sequencing and failure ownership look correct, and the focused tests cover the important boundaries. Before marking ready, please add or record an on-device smoke test for both successful replacement and forced configured-establish failure, confirming protected apps stall rather than bypass while excluded apps retain bypass; this behavior depends on Android’s `VpnService` interface handoff semantics and is not exercised by the pure unit test. Then mark ready and merge this one first.

PR #624, opened 2026-07-11
Make Insights blocked/allowed stat cards tap through to the detailed log

### LINKED ISSUE #563: From stats of blocked and unblocked trackers, allow to send to detailed log
**Is your feature request related to a problem? Please describe.**
No, just an improvement

**Describe the solution you'd like**
We should be able in the stats screen joined below to "click" some places to have more detailed stats on what are the blocked or unblocked trackers.

<details><summary>Screenshot</summary>
<p>

Image

</p>
</details>

kasnder:
## Summary

Closes #563.

The Insights screen (`InsightsActivity`) shows aggregate "Blocked" and "Allowed" tracker-attempt counts for the past 7 days, but there was no way to drill into which individual connections were blocked or allowed. This PR makes those two stat cards tappable so they navigate into the existing detailed traffic log (`ActivityLog`, the NetGuard-style per-connection log already reachable from the main menu), where users can see individual blocked/allowed connections.

## What changed

- `app/src/main/res/layout/activity_insights.xml`: gave the "Blocked" and "Allowed" stat card `LinearLayout`s ids (`llBlockedCard` / `llAllowedCard`), made them clickable/focusable with a ripple (`?android:attr/selectableItemBackground`), and added content descriptions for accessibility.
- `app/src/main/java/net/kollnig/missioncontrol/InsightsActivity.kt`: wires both cards to launch `ActivityLog` via a plain `Intent`, guarded by the same `Util.canFilter()` check (and `msg_unavailable` toast fallback) already used at the existing `ActivityLog` launch site in `ActivityMain.java`, so behavior is consistent across old/new Android versions.
- `app/src/main/res/values/strings.xml`: added two content-description strings (`insights_view_blocked_log`, `insights_view_allowed_log`).

## Investigation notes

- Found the stats screen described in the issue: the "Blocked"/"Allowed" cards in `activity_insights.xml`, rendered by `InsightsActivity`.
- Found the detailed log destination: `eu.faircode.netguard.ActivityLog` — the per-connection NetGuard-style log, already a standalone in-app `Activity` (declared in `AndroidManifest.xml`), currently launched only from `ActivityMain`'s `menu_log` item.
- Confirmed `ActivityLog` has no Intent-extra based pre-filter — its blocked/allowed/protocol filters are driven by persisted `SharedPreferences` toggled from its own options menu, not from an Intent extra a caller can set. Rather than repurpose those shared prefs as a side-channel filter (which would leave a persistent, possibly surprising, change to the user's log-filter settings), this PR just opens the log plainly, per the "keep scope tight" guidance — no new filtering subsystem was built.
- The per-app `TimelineFragment`/`TimelineAdapter` list was considered as an alternative destination, but it only shows tracker-company rollups per app, not the raw per-connection log the issue screenshot's counts refer to.

## Verification

- `./gradlew compileGithubDebugJavaWithJavac` — BUILD SUCCESSFUL (compiles both the Kotlin and Java sources, including the edited `InsightsActivity.kt`).
- Not manually verified on-device in this session (no Android environment attached to the sandbox); the change is a straightforward `OnClickListener` + `Intent` addition using patterns already present elsewhere in the codebase.

## Limitations

- Opens the detailed log without a pre-applied blocked/allowed filter, since `ActivityLog` doesn't expose an Intent-based filter mechanism (see investigation notes above). Users can still toggle "Show allowed"/"Show blocked" from the log's own options menu after arriving.

### DIFF
--- a/app/src/main/java/net/kollnig/missioncontrol/InsightsActivity.kt
+++ b/app/src/main/java/net/kollnig/missioncontrol/InsightsActivity.kt
@@ -46,6 +46,7 @@ import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.lifecycle.lifecycleScope
 import com.google.android.material.floatingactionbutton.FloatingActionButton
+import eu.faircode.netguard.ActivityLog
 import eu.faircode.netguard.Util
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -77,6 +78,8 @@ class InsightsActivity : AppCompatActivity() {
     private lateinit var llTopDomains: LinearLayout
     private lateinit var llNoData: LinearLayout
     private lateinit var llBlockedAllowed: LinearLayout
+    private lateinit var llBlockedCard: LinearLayout
+    private lateinit var llAllowedCard: LinearLayout
     private lateinit var progressBar: ProgressBar
     private lateinit var fabShare: FloatingActionButton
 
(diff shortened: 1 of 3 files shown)


kasnder:
Closing this — it satisfies the literal ask in #563 but not its intent, and the destination doesn't hold up:

1. **The traffic log is off by default.** `ActivityLog` reads the `log` DB table, which is only written when the `log` pref is on, and that defaults to `false` (`ServiceSinkhole.log()`: `prefs.getBoolean("log", false)` → `if (log) dh.insertLog(...)`). For nearly all users these cards would tap through to a blank screen.

2. **It's the wrong data source.** The "Blocked"/"Allowed" counts come from the `access` table (`InsightsDataProvider` → `getInsightsData7Days()` → `SELECT ... FROM access`), i.e. the per-app tracker log (`log_app`, default on). So the card summarizes `access` but links to a log built from a different, disabled table — the number and the destination aren't even the same data.

3. **No filter passthrough.** As the PR notes, `ActivityLog` has no Intent-based filter, so tapping "Blocked" opens the whole unfiltered log rather than the blocked connections — the link between the card and what's shown is lost.

4. `ActivityLog` is a NetGuard legacy per-connection screen, not part of TrackerControl's tracker-attribution model that the Insights numbers represent.

A meaningful drill-down for #563 should route to a tracker-attributed blocked/allowed breakdown backed by the same `access` table the counts come from, not to the legacy traffic log. Leaving #563 open for that.

Issue #231, opened 2021-07-20
Accessibility problem: unlabeled buttons!


PrivacyDragon:
Issue: The app list shows an app icon followed by it's name and sometimes some information thing. The app icon is a button. If you click on it, you disable internet access for that application. The problem is that the button is not labeled. A blind user can not hear what it is and what it does.

Fix: Add labels to the buttons. For example "Disable internet access for ..."

I might want to try to fix this myself if possible. If it is easy enough for me. It seems to me pretty easy, though sometimes I think something is easier than it actually is.

(I also posted this issue in my accessibility-hub repository: https://github.com/PrivacyDragon/FOSS-accessibility-hub/issues/7)

kasnder:
I agree that this is really poor practice. Thanks for raising this!

kasnder:
I've added it to the top of my todos.

kasnder:
What needs to be done is adding a label to this image: https://github.com/OxfordHCC/tracker-control-android/blob/3abc89d53b4d99596b0f62ed04ec855f42b3edda/app/src/main/res/layout/rule.xml#L12

kasnder:
Maybe, you could make a pull request, @PrivacyDragon ?

PrivacyDragon:
I will try.
Thanks for pointing out where the image is located in the code.

kasnder:
> I will try.
> Thanks for pointing out where the image is located in the code.

Great, thanks! Give it a try, otherwise I'm happy to look into it.

kasnder:
@PrivacyDragon Thanks for the PR. I've merged this. I'll leave this open for further investigation.

PR #625, opened 2026-07-11
Add dynamic accessibility labels to app-list Internet toggle (#231)

kasnder:
## What changed

Fixes the accessibility gap reported in #231: the app icon in the main app list is actually a button that blocks/allows Internet access for that app, but TalkBack had nothing meaningful to announce beyond a static, non-state-aware `contentDescription` added years ago (`tracked_app_icon`, still present as an inert XML default).

- **`app/src/main/java/eu/faircode/netguard/AdapterRule.java`**: added `updateToggleContentDescription()`, called on bind and again right after each tap, that sets the icon's `contentDescription` to reflect the app name and current state:
  - "Block internet access for {name}" when currently allowed
  - "Allow internet access for {name}" when currently blocked
  - "{name} has no internet access" when the app has no Internet permission at all (tapping is a no-op)
- **`app/src/main/res/values/strings.xml`**: added the three new string resources above, plus one more (see below), in the base (English) file only — translations are expected to follow via the usual localization process.
- **`app/src/main/java/net/kollnig/missioncontrol/details/TrackersListAdapter.java`** and **`app/src/main/res/layout/list_item_trackers_header.xml`**: extended the same fix to the per-tracker details screen, which has the same class of unlabeled-toggle problem:
  - the per-category "Block" `MaterialSwitch` now gets a dynamic content description naming its tracker category (`toggle_block_category_description`)
  - the header's Tracker Protection / Internet Access / Exclude from VPN switches (previously no accessible name at all beyond adjacency to a nearby TextView) now get their existing visible label text as `contentDescription`

No behavior/blocking-logic changes — this is purely accessibility labeling, per the issue's scope.

## Verification

- `./gradlew compileGithubDebugJavaWithJavac` — **BUILD SUCCESSFUL**.
- Not yet spot-checked on-device with TalkBack — please verify the announced text and that state updates correctly after a tap before merging.

Part of #231 (kept open as a tracking issue)

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/AdapterRule.java
+++ b/app/src/main/java/eu/faircode/netguard/AdapterRule.java
@@ -261,7 +261,9 @@ public class AdapterRule extends RecyclerView.Adapter<AdapterRule.ViewHolder> im
         // Show if Internet access blocked
         final ImageView iv = holder.ivIcon;
         InternetBlocklist internetBlocklist = InternetBlocklist.getInstance(context);
-        setGreyscale(iv, rule.internet && active && internetBlocklist.blockedInternet(rule.uid));
+        boolean blockedInternet = internetBlocklist.blockedInternet(rule.uid);
+        setGreyscale(iv, rule.internet && active && blockedInternet);
+        updateToggleContentDescription(iv, context, rule, blockedInternet);
         holder.ivIcon.setOnClickListener(view -> {
             if (!rule.internet)
                 return;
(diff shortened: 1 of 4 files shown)


kasnder:
Thanks—this addresses a real gap and CI is green. One correctness issue remains before merge: the app-icon description ignores `active`. If a UID is stored as blocked while tracker protection/apply is inactive, the UI renders it unblocked and tapping cannot toggle it (only `bypass_vpn_error`), but TalkBack announces “Allow internet access…”. Please derive the description from the effective/actionable state and announce the inactive case as unavailable (or disable that accessibility action). I would also prefer a neutral per-category label such as “Tracker blocking for %s” so the switch state communicates on/off. After that, spot-check with TalkBack and this should be ready.

kasnder:
Implemented in `7b8cd6c`: inactive Internet controls now announce as unavailable instead of exposing a false “Allow…” action, and category switches use neutral labels so TalkBack can communicate their checked state. `testGithubDebugUnitTest` and `assembleGithubDebug` pass locally. Remaining validation is an on-device TalkBack spot-check.

PR #626, opened 2026-07-11
Add Dependabot for automated dependency updates

kasnder:
## What

Adds `.github/dependabot.yml` to automate routine dependency bumps, and encodes the MaxMind `geoip2` version constraint as an enforced rule.

## Why

Two things wanted solving: automate regular dependency updates, and stop MaxMind `geoip2` from ever being bumped to 5.x (which uses Java constructs incompatible with Android). Dependabot handles both — its `ignore` rule is the natural home for the MaxMind constraint, turning a fragile version pin into an enforced, documented rule.

## Details

- **Gradle** (`/`) — weekly; minor/patch grouped into one PR to keep noise down, majors come individually for review.
- **Cargo** (`/wgbridge-rs`, the gotatun bridge) — weekly; sensitive, so treat these as informational PRs to test on-device before merging.
- **GitHub Actions** (`/`) — weekly; keeps `checkout`, `setup-java`, `gradle/actions`, `cache` current.
- **MaxMind constraint:**
  ```yaml
  - dependency-name: "com.maxmind.geoip2:geoip2"
    versions: [">=5.0.0"]
  ```
  Our usage is a single country-from-IP lookup (`details/CountriesFragment.java`), so there's nothing in 5.x worth a fork. 4.x patch/minor updates still flow, and **security** updates bypass version ignores — a transitive CVE in 4.x will still open a PR. An inline pin comment was added at the dependency so the rationale is visible in the build file.

## Notes

- Nothing auto-merges — every bump is a reviewed PR. Version updates activate once this merges to `master`; no repo toggle needed.
- Security updates require *Dependabot alerts* enabled in repo Settings → Code security (if not already on).


kasnder:
Superseded by #627, which carries the same Dependabot change cleanly based on `master` (this branch was based on an unmerged triage-doc commit that conflicted). Branch kept — it still holds the pending OPEN_ISSUES_TRIAGE.md update.

PR #642, opened 2026-07-12
Show app UID and shared-UID siblings, add uid: search filter (#229)

### LINKED ISSUE #229: Android groups applications on UID

kasnder:
## Summary

Closes #229.

Android apps can share a UID (e.g. via `android:sharedUserId`, or when the
system assigns the same UID to multiple packages). TrackerControl's
blocking is applied per-UID, so a setting change for one app silently
affects its shared-UID siblings — several users found this surprising,
and one commenter specifically asked for visibility into this without
adding UI clutter or new toggles.

This PR adds minimal, unobtrusive visibility into UID sharing:

- **Details screen** (`DetailsActivity`): shows the app's UID under the
  toolbar, and — only if other installed packages share that UID — a
  second line "Shares app ID with: ..." listing them, using
  `PackageManager.getPackagesForUid`.
- **App-list search** (`AdapterRule`'s `Filter`): supports a `uid:<number>`
  query to filter the list down to every app sharing a given UID, so users
  can go find the siblings mentioned in the details screen. Existing
  free-text/UID search behavior is unchanged.
- New strings (`app_uid`, `app_uid_shared_with`) added to the base
  English `strings.xml` only, per the project's translation workflow.

No new settings/toggles were added, keeping the UI as simple as before.

## Test plan

- [x] `./gradlew :app:compileGithubDebugJavaWithJavac` passes.
- [ ] **Manual/screenshot verification needed**: open an app's details
      screen and confirm the UID line appears (and the "shares app ID
      with" line appears for an app that has a shared-UID sibling, e.g.
      two apps from the same vendor using `android:sharedUserId`).
- [ ] **Manual/screenshot verification needed**: in the main app list,
      search `uid:<uid>` for a known UID and confirm only apps with that
      UID are shown.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/AdapterRule.java
+++ b/app/src/main/java/eu/faircode/netguard/AdapterRule.java
@@ -412,16 +412,21 @@ public class AdapterRule extends RecyclerView.Adapter<AdapterRule.ViewHolder> im
                     listResult.addAll(listAll);
                 else {
                     String queryStr = query.toString().toLowerCase().trim();
+                    // Apps sharing a UID are controlled together by the VPN (#229), so let
+                    // users explicitly search for them all with a "uid:<number>" query.
+                    boolean uidOnly = queryStr.startsWith("uid:");
+                    String uidStr = uidOnly ? queryStr.substring(4).trim() : queryStr;
                     int uid;
                     try {
-                        uid = Integer.parseInt(queryStr);
+                        uid = Integer.parseInt(uidStr);
                     } catch (NumberFormatException ignore) {
                         uid = -1;
                     }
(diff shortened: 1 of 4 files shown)


kasnder (commented):
Reviewed the UID display, shared-package lookup, `uid:` filter behavior, layout integration, and surrounding details/search code. I found no actionable correctness issue. `:app:compileGithubDebugJavaWithJavac` succeeds, and the branch is clean.

Issue #216, opened 2021-06-24
Temporary unblock


dangavin:
I would like to have the ability to temporarily unblock a service (say, for instance, by long pressing the service on the Manage App screen).

The use-case for this arises as I would sometimes like to use apps and websites that have a dependency on services I would generally block, but which, in these particular cases, I would allow. By *temporarily* unblocking the service, I would not need to remember to block it again later.

I would think a ten minute timeout on the temporary unblocking would suffice, but perhaps this could be configurable.

kasnder:
Thanks for the suggestion! What you can already do is pause TrackerControl for a few minutes from its notification.

I agree that the better solution would be to have this on a per-app basis.

nmarteybotchway:
I would be interested working on this

kasnder:
@nmarteybotchway i think this requires too extensive changes to the code base.

PR #643, opened 2026-07-12
Temporarily allow a blocked tracker per app (#216)

kasnder:
Implements #216: temporarily unblock a blocked tracker company/service for a single app, so a user can complete a flow that a blocked tracker breaks, without having to remember to re-block. Per the maintainer's note on the issue, this is done **per-app** (rather than the existing global pause via the notification).

## What it does
- In an app's tracker details list, **long-pressing a BLOCKED tracker company row** shows a dialog offering **"Allow for 10 minutes"**.
- On confirm, the company is unblocked immediately for that app and an alarm is scheduled to re-block it after 10 minutes.
- While active, the row shows **"ALLOWED FOR 10 MIN"**. Manually re-blocking the row (tap) ends the allowance early and cancels its alarm.

## Design decisions
- **Simplicity over configurability** (project philosophy): a single fixed 10-minute duration, no settings screen, no per-duration UI.
- **No reload needed for the unblock itself.** The packet path reads the live `TrackerBlocklist` singleton per connection (`ServiceSinkhole` ~line 2540), so `unblock()` + `saveSettings()` take effect immediately for new connections — matching the existing per-tracker tap toggle, which also does not reload.
- **Re-block does reload defensively.** When the alarm fires (or the sweep re-blocks), `TempUnblockReceiver` calls `Rule.clearCache()` + `ServiceSinkhole.reload(...)` to guarantee the change reaches the packet path, mirroring the header toggles.
- **Alarm** mirrors `WidgetAdmin`'s pause alarm: `setAndAllowWhileIdle(RTC_WAKEUP, …)`, allow-while-idle so it can fire in Doze, no `SCHEDULE_EXACT_ALARM` permission. Inexact, so under deep Doze the re-block may lag slightly (acceptable — the fail-safe direction is re-blocking late, not never).
- **Process-death safety.** Expiry timestamps are persisted in a `temp_unblock` SharedPreferences file. `ServiceSinkhole.onCreate()` calls `TempUnblockReceiver.sweep()`, which re-blocks anything already expired and reschedules alarms for anything still active. So a killed process cannot leave a tracker permanently unblocked.
- New strings are English-only in `values/strings.xml`; translations flow via Crowdin.

## Files
- `app/src/main/java/eu/faircode/netguard/TempUnblockReceiver.java` (new) — allowance state, alarm scheduling, sweep, re-block.
- `TrackersListAdapter.java` — long-press handler, "ALLOWED FOR 10 MIN" status, cancel-on-manual-reblock.
- `ServiceSinkhole.java` — sweep on service start.
- `AndroidManifest.xml` — registers the (non-exported) receiver.
- `values/strings.xml` — new user-facing strings.

## Needs on-device verification
- Alarm actually fires and re-blocks under Doze (allow-while-idle timing).
- Re-block reload reaches the packet path (a previously-allowed tracker is blocked again after 10 min without reopening the app).
- Sweep behaviour after force-stop / reboot with an allowance pending.
- Long-press UX on the `ListView` rows across supported API levels.

## Note
A community contributor volunteered on the issue on 2026-07-12. This draft is a **starting point** — the maintainer may prefer to hand it over to that contributor or use it as a reference.


kasnder (commented):
Review finding: `TempUnblockReceiver.java` contained two literal NUL bytes in Java string/character literals used as the SharedPreferences key separator. NUL makes the source invalid Java input (and is not a safe XML-backed preference-key delimiter), so this prevented the new receiver from compiling. I replaced the separator with `:` in commit 1bbf1c1 and verified `:app:compileGithubDebugJavaWithJavac` succeeds.

PR #647, opened 2026-07-12
Add tracker:<company> app-list search (#447)

kasnder:
Implements #447: let users search which installed apps have contacted a particular tracker company or SDK.

## What

The existing app-list search box now understands a `tracker:` prefix. A query like

```
tracker:gravy analytics
```

filters the app list to apps whose observed/contacted tracker companies contain the query substring (case-insensitive). Plain queries keep the existing name/package/uid search behaviour unchanged.

## How

- `TrackerList.getTrackerNamesByUid()` — a single scan over the recorded hosts, mapping each app UID to the set of contacted tracker company names (same traversal pattern as the existing `getTrackerCountsAndTotal()`).
- `AdapterRule.getFilter()` — on a `tracker:`-prefixed query, builds that lookup once per filter invocation on the `Filter`'s background thread (no per-row DB queries, no UI-thread work), then keeps apps whose company set has a substring match.
- Discoverability: the SearchView now shows the query hint "Search app, or tracker:company" (new base-English string `menu_search_hint` only; translations left to Crowdin).

No new screens, settings, or dependencies.

## Usage example

1. Open TrackerControl, tap the search icon on the Apps tab.
2. Type `tracker:facebook` — the list shrinks to apps that have contacted Facebook-owned tracker domains.
3. Tap an app to open its details and see/block the trackers.

## Manual verification needed

- On a device with recorded traffic: `tracker:facebook` (or another company seen in app details) returns the expected apps; a nonsense company returns an empty list.
- Plain searches (app name, package, numeric uid) behave exactly as before.
- The new query hint shows in the expanded SearchView and does not truncate awkwardly on narrow screens.
- Search performance with a large host log (lookup is one DB scan per keystroke after the 300 ms debounce — please sanity-check on a long-running install).

Note: like the "contacted N companies" counts, this searches *observed* tracker contacts recorded by TC, not static SDK analysis, and interacts with the hidden-app search stubs from 58a9e0a only in that tracker-search matches are restricted to the visible list (stubs carry no tracker data).

Closes #447

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityMain.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityMain.java
@@ -911,6 +911,9 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
         });
 
         final SearchView searchView = (SearchView) menuSearch.getActionView();
+        // Hint at the "tracker:<company>" search prefix (#447), which filters apps
+        // by the tracker companies they have been observed contacting.
+        searchView.setQueryHint(getString(R.string.menu_search_hint));
         searchView.setOnQueryTextListener(new SearchView.OnQueryTextListener() {
             @Override
             public boolean onQueryTextSubmit(String query) {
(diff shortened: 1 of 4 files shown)


kasnder:
**Review finding — locale-independent matching** (`AdapterRule.java`): The case-insensitive comparison depends on the device locale. In Turkish, for example, `"Branch Metrics".toLowerCase()` contains a dotless `ı`, so `tracker:metrics` will not match despite the feature promising case-insensitive search. Both the query and tracker name should be normalized with `Locale.ROOT`. I have prepared that fix on the PR branch.

Issue #652, opened 2026-07-12
Profile and reduce Secure DNS screen-off battery use


kasnder:
## Problem

Secure DNS currently uses the Java-based local DNS proxy / DoH client and may keep devices warm while the screen is off. It should not become a stronger default until idle behavior is understood and fixed.

## Investigate

- Whether the local DNS proxy or DoH client remains active without DNS traffic
- Wakeups from retries, circuit-breaker checks, network changes, or idle HTTPS connections
- DNS-cache effectiveness and repeated upstream DoH queries
- Duplicate work or conflicts with WireGuard-provided DNS
- Whether routing system apps through TrackerControl/DoH contributes to wakeups

## Desired direction

After the screen-off cost is low, add a simple protection-health screen for tracker blocking, Secure DNS, WireGuard, Android Private DNS conflicts, and battery/background-permission status.

Keep the defaults simple: low-battery tracker blocking by default, with Secure DNS presented as an explained stronger-privacy option until its idle cost is acceptable. Avoid expert configuration unless it directly helps users recover from breakage.

PR #657, opened 2026-07-12
Reduce Secure DNS screen-off battery use: RFC 8484 caching + idle-socket eviction

kasnder:
Part of #652.

### RFC 8484 / HTTP-native response caching

The DoH client now uses cache-friendly GET requests for normal DNS messages and a bounded 2 MiB OkHttp cache. Before sending upstream it normalizes the DNS transaction ID to zero, as recommended by RFC 8484, so otherwise-identical queries share one HTTP cache key. The complete remaining wire message stays in the URL, so flags, query casing, DNSSEC state, and EDNS options remain part of the key.

Freshness comes from the DoH server HTTP response (`Cache-Control` / `Age`) instead of a TrackerControl-owned DNS cache policy. On a cached response, dnsjava parses and serializes the message, restores the local callers transaction ID, and subtracts the HTTP response age from resource-record TTLs as RFC 8484 requires. OPT metadata is preserved rather than treated as a TTL.

Queries whose GET URL would exceed 2048 characters fall back to POST. They still use transaction ID zero but intentionally bypass HTTP caching.

This replaces and deletes the original hand-written `DnsResponseCache` implementation and its DNS wire-format walker. The production dependency is dnsjava 3.6.4 (588 KiB before R8, no added transitives in the resolved app configuration).

The default Quad9 endpoint was checked directly and returns cacheable DoH GET responses (`Cache-Control: max-age=...`). Custom RFC 8484 endpoints control their own cacheability through normal HTTP headers.

### Screen-off idle-socket eviction

On `ACTION_SCREEN_OFF`, `ServiceSinkhole` calls `DnsProxyServer.onScreenOff()`, which evicts idle OkHttp keep-alive connections on the existing background executor. In-flight requests are unaffected and the HTTP response cache remains available while the screen is off.

The proxy still clears the HTTP cache when it stops or changes network context, and existing WireGuard DNS handling remains unchanged.

## Why this shape

The native C DNS parser remains responsible for TrackerControl packet-path attribution and blocking. The DoH client does not duplicate that parser or attempt to maintain a second resolver/cache implementation. OkHttp owns HTTP caching and dnsjava owns standards-sensitive DNS parsing/serialization.

## Verification

- `./gradlew :app:testGithubDebugUnitTest` — passes.
- Focused tests cover ID-zero normalization, equivalent HTTP cache keys across caller IDs, complete-message cache-key separation, unpadded base64url encoding, GET/POST selection, transaction-ID restoration, TTL aging, and TTL underflow protection.
- `./gradlew :app:compileGithubDebugJavaWithJavac -q` — passes.

## On-device follow-up

Measure a fixed screen-off window with DoH enabled and no WireGuard egress. Compare radio-active time and TC-attributed wakeups before/after, and inspect `TrackerControl.DoH` cache-hit logs. No Secure DNS default is changed by this PR.


kasnder:
Review finding fixed in `28cc9da` (pushed to this PR).

The response cache previously imposed a 30-second minimum TTL. That makes it serve a DNS answer after its authoritative TTL has expired (for example, a 5-second record for another 24 seconds), which can retain a stale destination and is not DNS-correct. The cache now only caps excessively long TTLs; the focused test verifies that the 5-second record expires after 5 seconds.

Focused `DnsResponseCacheTest` passes. The screen-off idle-socket eviction and the bounded positive-response cache otherwise look aligned with the battery goal.

kasnder:
Follow-up implementation pushed in `d38c337` and `3aa2201`.

The hand-written DNS response cache and parser are now deleted. Normal queries use RFC 8484 GET with ID zero and OkHttp HTTP caching; oversized requests fall back to POST. dnsjava handles the one DNS-aware operation HTTP caching cannot do itself: restoring the caller ID and subtracting HTTP response age from every real RR TTL before returning a cached message. This avoids extending downstream TTLs while preserving EDNS OPT metadata.

The full GitHub-flavour unit suite passes. I also verified that the default Quad9 endpoint serves GET responses with `Cache-Control: max-age=...`.

kasnder:
Release-size follow-up: dnsjava 3.6.4 is 588 KiB as a raw JAR (317 classes). The release build has `minifyEnabled = true`; R8 retains 120 dnsjava classes and removes the other 197. Many record decoders remain reachable through dnsjava’s type registry, so it is shrunk substantially but not reduced to only the four directly imported classes.

The first release build also exposed dnsjava’s optional desktop-JVM SPI/logging references. Commit `21c2695` adds narrowly scoped `dontwarn` rules for those unused Android-inapplicable paths. `:app:assembleGithubRelease` now succeeds; the resulting universal four-ABI APK is 13,810,639 bytes. An exact dependency-only APK delta still needs an otherwise-identical baseline build.

kasnder:
## Code review summary

Reviewed the RFC 8484 caching + idle-socket eviction changes. Solid, well-tested change with sound architecture (OkHttp owns HTTP caching, dnsjava owns DNS serialization, the native C parser stays authoritative). Verified the correctness-critical bits:

- **ID normalization/restoration** is right and necessary — zeroing upstream but restoring the caller's ID before return so the proxy can match responses. Copy-not-mutate confirmed.
- **Screen-off trigger** is correct: `!last_interactive` genuinely means screen-off, and `onScreenOff()` is a safe no-op when the proxy isn't running.
- **No dangling callers** of the old `getInstance(String)` signature.
- **Cache key**: full wire message (minus ID) — no cross-query collision; flags/EDNS/qname-case preserved. RFC 8484 compliant.

### Issues found and fixed in b62c246f

1. **dnsjava was on the every-response path (robustness regression).** `resolve()` re-parsed and re-serialized *every* response via dnsjava purely to restore the transaction ID. Any parse/round-trip quirk turned a previously-working resolution into `null` (retries just re-fetch the same cached bytes and fail identically). Fixed: fresh responses (`age == 0`, the common path) now restore the ID by patching the two ID bytes directly, keeping dnsjava off the hot path. The full round-trip runs only when TTLs actually need aging, and if aging fails the answer is still returned with its ID restored instead of dropped.
2. **Redundant double `normalizeTransactionId`** in `buildRequest` — removed.
3. **`buildRequestUrl`** built a throwaway `Request` just to get an `HttpUrl` — now uses `HttpUrl.parse` with an invalid-endpoint guard.
4. Added tests for the new fresh-response fast path (10/10 pass).

### Release build (R8) — verified
The PR noted the minified release path was untested. I ran `assembleGithubRelease`: dnsjava's record classes survive R8 shrinking **without** a keep rule, so I deliberately avoided a broad `-keep class org.xbill.DNS.**` (it would bloat the APK against the "588 KiB before R8" intent). Only remaining noise was dnsjava's desktop-JVM `InetAddressResolver` SPI dangling in `META-INF/services`; one `-dontwarn` silences it (complements the JVM-service dontwarns in 21c26955).

### Left as judgment calls (no change)
- `rdataToWireCanonical()` lowercases embedded domain names in aged records — functionally harmless (DNS is case-insensitive), but cached answers can differ byte-for-byte from what the server sent. `rdataToWire()` would preserve original casing if preferred.
- DoH answers are now persisted to disk (`cacheDir/doh-http-cache`) vs. the previous in-memory cache — a mild change in resident data for a privacy tool. Cleared on shutdown/network change, so acceptable; worth a line in the PR description.

Issue #655, opened 2026-07-12
Decide how to reduce cross-app ambiguity in DNS-based tracker blocking


kasnder:
## Current behavior

The access table has a UID, destination IP, block decision, and uncertainty flag. DNS observations are global: they store qname, aname, resource, time, and TTL, but no UID. Consequently, `getQAName(uid, ip, ...)` cannot determine which hostname a specific app resolved; it reuses hostnames globally observed for that IP.

This can cause both shared-IP ambiguity and cross-app attribution ambiguity in per-app blocking decisions.

## Options

1. Low-risk partial improvement: key the in-memory tracker-verdict cache by `(uid, destination IP)` instead of destination IP alone. This reduces cross-app cache bleed but does not make DNS evidence app-specific.
2. Full attribution redesign: persist DNS observations with a UID, link DNS evidence to access observations, or adopt a different attribution model. This affects schema, migration, collection/query logic, and the semantics of `uncertain`.

## Decision needed

Choose whether the product goal is only more-conservative cache reuse or genuine per-app DNS evidence, and decide how conservative standard blocking should be when evidence remains global and ambiguous.

kasnder:
#654 found technical gaps in current tls.c
No cross-segment reassembly: a ClientHello split across TCP segments (increasingly common — large extension sets, post-quantum key shares push past one MTU) is silently missed, and checkedHostname means no retry.
Single TLS record only.
Memory safety: parser is bounds-checked throughout; no OOB found (one harmless cosmetic off-by-one at tls.c:66).
Cost: the parse itself is negligible, but the control-flow change moves verdict work from the cheap SYN path onto the hot data path for port 443, and robust reassembly would add per-connection buffering — against the battery-first constraint.

PR #658, opened 2026-07-12
Key tracker-verdict cache by (uid, destination IP) — partial #655

kasnder:
Part of #655.

## What this does

Implements **Option 1** from #655: the in-memory tracker-verdict caches in `ServiceSinkhole` (`ipToHost`, `ipToTracker`) are now keyed by a composite **(uid, destination IP)** via `trackerCacheKey()` instead of destination IP alone.

These caches sit directly on the per-app blocking path (`blockKnownTracker(daddr, uid)`). With an IP-only key, the first app to reach a shared IP populated a single cache slot whose verdict was then reused for every other app hitting that IP. Per-app keying makes cache reuse conservative and removes that cross-app cache bleed.

A freshly resolved DNS answer for an IP must drop **all** apps' cached verdicts for that IP, so `dnsResolved()` now calls `invalidateTrackerCachesForIp()`, which prefix-scans the maps on `"<ip>#"` (`#` never appears in an IP literal).

`Expiring<T>` is promoted from a private inner class to a package-private `static` nested class (it uses no outer state) so the new keying and invalidation can be unit-tested.

### Caches reviewed

| Cache | Key before | Change | Why |
|---|---|---|---|
| `ServiceSinkhole.ipToTracker` / `ipToHost` | destination IP | **now (uid, IP)** | On the per-app blocking path; must not bleed across apps. |
| `ServiceSinkhole.mapUidIPFilters` | `IPKey(version, protocol, dport, uid)` | left as-is | Already uid-keyed. |
| `TrackerList.hostnameToTracker` | hostname | left global | A hostname belongs to a tracker regardless of app — genuinely uid-independent. |
| `get_uid` cache in native `ip.c` | 5-tuple | left as-is | Resolves the uid itself; not a verdict cache. |

Note: the native `wg_flow_cache` referenced in the issue-triage notes does **not** exist in the current tree (WireGuard-mode passive DNS in `wgbridge-rs` never caches a block/allow verdict), so there was nothing to fix there.

### Cost

Cache cardinality scales with **IPs × active apps** instead of IPs. Entries still expire on the DNS TTL (or the 60 s negative-cache TTL for misses), and the maps are fully cleared on the periodic mapping refresh, so the working set stays bounded by live traffic rather than growing unbounded.

### Honest scope note

Because `getQAName(uid, ip, ...)` still ignores `uid`, the *value* derived for a given IP is currently the same for every app, so this change has no immediate effect on which apps get blocked today. Its purpose is structural: it removes cache-level cross-app reuse and guarantees no cross-app bleed the moment any uid-specific state (per-app access-table verdicts, or a future uid-aware DNS lookup) feeds the cached value. That is exactly the "more-conservative cache reuse, not app-specific evidence" framing of Option 1.

## Verification

- `./gradlew :app:compileGithubDebugJavaWithJavac -q` — passes.
- `./gradlew :app:testGithubDebugUnitTest` — **141 tests, 0 failures**, including the new `TrackerCacheKeyTest` (per-uid key distinctness, IP-prefix invalidation, and that invalidating one IP drops all apps for that IP while sparing other IPs).
- No native C touched, so no `assembleGithubDebug` needed.

## Decision input for #655

**Recommendation: ship Option 1 now (this PR); do not pursue full Option 2 yet — instead make Standard mode's ambiguous-IP handling explicit, and only revisit per-app DNS attribution if real reports show it is needed.**

### What Option 2 would require

Option 2 ("genuine per-app DNS evidence") means persisting DNS observations with a UID and linking DNS evidence to access observations. Concretely:

1. **Schema + migration.** Add a `uid` column to the `dns` table and re-key/re-index it. The `dns` table is high-churn and can be large; an online migration that rewrites it (plus its `resource` index) is a non-trivial, risky one-off on-device cost. Historical rows have no uid and would need a sentinel (`-1`/global), so queries must handle mixed uid/global rows indefinitely.
2. **Collection path.** DNS is parsed in native C (`dns.c`) / the Rust bridge and recorded via `dnsResolved()` / `wireGuardDnsResolved()`, none of which currently carry a reliable uid at DNS-record time. The querying app for a DNS response is frequently a shared resolver/system UID, not the app that will later use the IP — so attributing the answer to a uid at capture time is often wrong. This is the core reason DNS was left uid-global in the first place.
3. **Query logic + `uncertain` semantics.** `getQAName`, the `log()` path, and `blockKnownTracker` would all need uid-aware variants, and the meaning of `ACCESS_UNCERTAIN_SHARED_IP` / `MIXED` / `MULTIPLE_TRACKERS` would change: today "uncertain" captures *shared-IP-across-apps* ambiguity; under per-app evidence it would need to distinguish "this app resolved a tracker hostname for this IP" from "some app did". That is a semantic redesign of the insights/blocking contract, not a localized change.
4. **Attribution model mismatch.** Per AGENTS.md §4, attribution is deliberately **global, not per-app** — a known, deferred limitation. Option 2 reverses a standing design decision and would need product sign-off, not just an engineering change.

### Why not now

- **Cost/benefit.** Option 2 touches schema, migration, native + Rust collection, query logic, and UI semantics for a gain that is only real when an IP is genuinely shared by a tracker host and a non-tracker host used by *different* apps — a narrow slice already partly mitigated by the Minimal/Standard/Strict ambiguous-IP handling.
- **Correctness risk.** The capture-time uid is often the wrong uid (shared resolver UIDs), so naive Option 2 could make attribution *less* accurate while looking more precise.
- **Reversibility.** Option 1 is additive and reversible; Option 2's schema/migration is not.

### The real product decision the issue asks for

The issue also asks "how conservative Standard blocking should be when evidence is global and ambiguous." That is orthogonal to Option 2 and cheaper to get right: it is already expressed by `BlockingModeLogic.blocksAmbiguousTrackerIp()` (Standard allows ambiguous shared-IP hosts, Strict blocks them). If cross-app ambiguity is causing real breakage or leakage reports, the low-cost lever is to tune that policy (and document it), not to rebuild DNS attribution. I'd recommend treating that policy tuning as the primary follow-up, and keeping full per-app DNS attribution (Option 2) deferred behind a concrete, reported need.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -2207,8 +2207,7 @@ public class ServiceSinkhole extends VpnService {
             prepareUidIPFilters(rr.QName);
 
             if (Util.isNumericAddress(rr.Resource)) { // make sure correct format
-                ipToHost.remove(rr.Resource);
-                ipToTracker.remove(rr.Resource);
+                invalidateTrackerCachesForIp(rr.Resource);
             }
         }
     }
@@ -2257,8 +2256,35 @@ public class ServiceSinkhole extends VpnService {
 
     static ConcurrentHashMap<Integer, String> uidToApp = new ConcurrentHashMap<>();
     private static final ConcurrentHashMap<Integer, String> uidToPackage = new ConcurrentHashMap<>();
+    // The tracker-verdict caches are keyed by (uid, destination IP) via
+    // trackerCacheKey(), not destination IP alone (see #655, Option 1). These
+    // caches sit directly on the per-app blocking path
+    // (blockKnownTracker(daddr, uid)); an IP-only key lets one app's cached
+    // verdict be reused for a different app that reaches the same shared IP.
+    // Per-app keying keeps cache reuse conservative and blocks cross-app bleed
+    // the moment any uid-specific state (per-app access rules, or a future
+    // uid-aware DNS lookup) feeds the cached value. Cost: cache cardinality
+    // scales with IPs x active apps instead of IPs; entries still expire on the
+    // DNS TTL (or NEGATIVE_TRACKER_CACHE_TTL_MS for misses), so the working set
+    // stays bounded by live traffic.
     static ConcurrentHashMap<String, Expiring<String>> ipToHost = new ConcurrentHashMap<>();
     static ConcurrentHashMap<String, Expiring<Tracker>> ipToTracker = new ConcurrentHashMap<>();
+
+    // Composite (uid, destination IP) key for the tracker-verdict caches. The
+    // '#' separator never appears in an IPv4/IPv6 literal, so the IP prefix
+    // stays unambiguous for invalidateTrackerCachesForIp().
+    static String trackerCacheKey(int uid, String daddr) {
+        return daddr + '#' + uid;
+    }
+
+    // A freshly resolved DNS answer for an IP invalidates every app's cached
+    // verdict for that IP, so drop all (uid, IP) entries sharing this IP.
+    static void invalidateTrackerCachesForIp(String daddr) {
+        String prefix = daddr + '#';
+        ipToHost.keySet().removeIf(k -> k.startsWith(prefix));
+        ipToTracker.keySet().removeIf(k -> k.startsWith(prefix));
+    }
+
     static String NO_DNAME = "null"; // use a String, unequal the real null
     static Tracker NO_TRACKER = new Tracker(null, null, 0);
     // Negative results (no tracker / no dname for an IP) are cached only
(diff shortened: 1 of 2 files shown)


kasnder:
Request changes: this PR should not merge in its current form.

The PR correctly documents that `getQAName(uid, ...)` still ignores the UID, so this cache split has no current effect on tracker detection or blocking. It does, however, multiply cache cardinality by active apps and makes every DNS invalidation scan all per-UID entries for the IP.

More importantly, the repository guidance explicitly treats DNS attribution as deliberately UID-global and deferred; it says not to patch that limitation ad hoc. A UID-keyed verdict cache is neither reliable UID evidence nor a user-visible breakage recovery. Please close this partial change and revisit only with a product-approved, end-to-end attribution design backed by a concrete report.

PR #692, opened 2026-07-28
Address review findings on relay failover (PR #688)

kasnder:
## Summary

Builds on #688 ("Fail over to a different Mullvad/IVPN relay when the current one stays dead") with a follow-up commit addressing review findings on that change:

- Bound the relay-list fetch with a real timeout (`FAILOVER_TIMEOUT_MS`): previously `forceRestartPending` blocked every other recovery path while the unbounded network call ran, so a hang could stall recovery longer than not having failover at all.
- Made `failoverInFlight` an `AtomicBoolean` to close a check-then-set race between the monitor and rebind threads.
- Only trigger failover from monitor-observed breaks, not from rebind failures after a network change (those aren't evidence the relay is dead).
- Reset the restart backoff after a successful switch so a fresh relay doesn't inherit the old one's penalty.
- Persist the IVPN session `WgRelayFailover` generates/reuses; it was minting sessions that never got saved, orphaning them against the account's device limit and desyncing stored session state from the live config.
- Scoped the config write to the profile ID resolved before the network call (`WgProfileManager#updateProfileConfigIfActive`), so a profile switch mid-fetch can't overwrite a different profile's stored config.
- Excluded the relay that just failed from the next pick when another candidate exists, and reject a pick that silently widens the exit country, for both Mullvad and IVPN.
- Distinguished `ApiRejectedException`/`CaptchaRequiredException` from other failures in the failover log.
- Added `WgRelayFailoverTest` and extended `WgProfileManagerTest` for the new guard clauses.

## Test plan

No Android SDK is available in this sandbox, so the change could not be compiled/tested locally; please run CI.

- [ ] `./gradlew :app:compileGithubDebugJavaWithJavac`
- [ ] `./gradlew :app:testGithubDebugUnitTest`
- [ ] Manually verify: with a Mullvad/IVPN WireGuard profile active, block the current relay's IP and confirm TC switches to a different relay (excluding the dead one, without widening country) after a few restart cycles, and that the switch doesn't stall recovery if the relay-list fetch hangs.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mf2a5cVzdZ6ovMAVfmCoyg)_

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -1807,7 +1807,8 @@ public class ServiceSinkhole extends VpnService {
         // It also handles the disable case internally — no need to gate.
         net.kollnig.missioncontrol.wg.WgEgress.INSTANCE.setRecoveryCallbacks(
                 () -> ServiceSinkhole.reload("wireguard connectivity repair", ServiceSinkhole.this, false),
-                () -> showWireGuardErrorNotification(getString(R.string.msg_wg_recovery_failed)));
+                () -> showWireGuardErrorNotification(getString(R.string.msg_wg_recovery_failed)),
+                () -> net.kollnig.missioncontrol.wg.WgRelayFailover.attemptFailover(ServiceSinkhole.this));
         jni_wireguard_required(prefs.getBoolean("wg_enabled", false)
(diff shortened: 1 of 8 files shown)


kasnder:
Closing — this was opened as a standalone follow-up before realizing the fix commit should just land on #688's own branch (`claude/github-issue-679-2zbcql`) instead. Pushed there directly; see #688.

---

PR #778, opened 2026-08-22
Fix IPv6 extension header walk in handle_ip()

### LINKED ISSUE #769: Correctly walk IPv6 extension headers in handle_ip() (ip.c)
## Background

PR #751 ("Fix memory-safety and hang bugs in the native packet engine") originally included a hunk that made `handle_ip()` in `app/src/main/jni/netguard/ip.c` actually walk IPv6 extension headers to find the upper-layer protocol. It was pulled out of that PR before merge because review found it unsafe. This issue tracks doing the walk correctly.

## Current (master) behavior

master's loop at `ip.c:362-385` looks like:

```c
uint16_t off = 0;
protocol = ip6hdr->ip6_nxt;
if (!is_upper_layer(protocol)) {
    off = sizeof(struct ip6_hdr);
    struct ip6_ext *ext = (struct ip6_ext *) (pkt + off);
    while (is_lower_layer(ext->ip6e_nxt) && !is_upper_layer(protocol)) {
        protocol = ext->ip6e_nxt;
        off += (8 + ext->ip6e_len);
        ext = (struct ip6_ext *) (pkt + off);
    }
    if (!is_upper_layer(protocol)) {
        off = 0;
        protocol = ip6hdr->ip6_nxt;
    }
}
payload = (uint8_t *) (pkt + sizeof(struct ip6_hdr) + off);
```

The loop condition tests `is_lower_layer(ext->ip6e_nxt)` (the *next* header's type) but the body assigns `protocol = ext->ip6e_nxt`. So by the time the loop exits, `protocol` always holds a lower-layer value — the `if (!is_upper_layer(protocol))` fallback right after the loop always fires, resets `off` to 0, and `payload` lands right after the fixed IPv6 header, on top of whatever extension header is actually there. In other words: **master never successfully dispatches a chained-extension IPv6 packet to TCP/UDP today; `off` is effectively dead code.** This is not a regression to fix — it's the reason the PR's naive fix (just changing the loop condition to `is_lower_layer(protocol)`) looked like a strict improvement while actually introducing three new bugs, below.

## Why the straightforward fix is unsafe

Changing the loop condition to `is_lower_layer(protocol)` makes `off` live for the first time, which exposes:

1. **Wrong unit for the header length.** The PR did `off += (8 + ext->ip6e_len)`. Per RFC 8200, `Hdr Ext Len` is in 8-octet units *not counting* the first 8 octets, so the correct advance is `8 * (ext->ip6e_len + 1)`. Any extension header bigger than the minimum 8 bytes lands `payload` inside the header instead of past it. The PR's added bounds check (`off + 8 + ext->ip6e_len > length`) inherits the same wrong unit, so it doesn't actually protect against the miscalculation.

2. **ESP and AH cannot be walked as `struct ip6_ext`.** `is_lower_layer()` includes ESP (protocol 50), which has no next-header field at all — the walk would read the SPI's first byte as `ip6e_nxt`, garbage. AH (protocol 51) does have a next-header field, but its length field is in 4-octet units minus 2, not the `struct ip6_ext` 8-octet-minus-8 convention, so reusing the same struct/arithmetic misparses it too.

3. **Non-first fragments must not be dispatched.** Fragment (protocol 44) would now be walked, but a non-first fragment has no L4 header at all — its payload bytes are arbitrary earlier data, and parsing them as a TCP/UDP header creates a bogus session and can emit packets to a random host/port. IPv4 already rejects fragments outright (`ip4hdr->frag_off & IP_MF` check at `ip.c:332-336`); IPv6 has no equivalent guard.

## Suggested shape for a real fix

- Restrict the walk to the extension header types that are actually `struct ip6_ext`-shaped and safe to skip: Hop-by-Hop (0), Destination Options (60), Routing (43), and possibly Mobility (135) — using the correct `8 * (ext->ip6e_len + 1)` advance.
- Stop the walk (do not dispatch) on ESP (50) or AH (51) — there is no generic, safe way to skip past encrypted/authenticated payloads with the current per-flow model.
- Treat Fragment (44) as a fixed 8-byte header (`struct ip6_frag`) and only continue/dispatch when it is the *first* fragment (fragment offset == 0); otherwise bail out the same way IPv4 does for `IP_MF`/non-zero offset.
- Keep a length check before every dereference of the extension header (bounded by `length`, using the corrected unit).

## Related

- The bounded-but-wrong version was reverted out of #751 in favor of leaving master's (already broken but at least inert) behavior in place, plus adding an `ihl`/`doff`-style validation guard where relevant elsewhere in `ip.c`.

kasnder:
Fixes #769.

## User-visible impact today (before this fix)

The extension-header walk in `handle_ip()` has a loop-termination bug: the `while` condition tests `is_lower_layer(ext->ip6e_nxt)` (the header *after* the one currently being examined) but the code assigns `protocol = ext->ip6e_nxt` — so `protocol` can only ever be re-assigned another *extension-header* type, never the real upper-layer protocol found at the end of the chain. Concretely, even the simplest real-world case — a single Hop-by-Hop header immediately followed by TCP — fails: the loop body never executes at all (`ext->ip6e_nxt` there already is TCP, which is not "lower layer"), so the walk always falls through to the bail-out branch, which resets `off = 0` and `protocol = ip6hdr->ip6_nxt` (the *first* extension-header type, e.g. `0`/Hop-by-Hop).

Downstream, `protocol` no longer matches `IPPROTO_TCP/UDP/ICMP/ICMPV6`, so:
- **`sport`/`dport` stay 0** and none of the TCP/UDP/ICMP-specific fast paths apply; the packet instead goes through the generic IP-based `is_address_allowed()` block decision (so IP/tracker blocking itself still works, just without any port-aware nuance).
- If the destination is **allowed** and the flow is **not** tunnelled through WireGuard, none of `handle_icmp`/`handle_udp`/`handle_tcp` fire (none match `protocol`), so **the packet is silently dropped** — never forwarded, and not logged as blocked either (this is a connectivity bug, not a privacy one).
- If the destination is **allowed** and the flow **is** tunnelled through WireGuard, `write_wireguard_packet()` forwards the full raw packet byte-for-byte regardless of `protocol`/`payload`, so **WireGuard-routed IPv6 traffic with extension headers is unaffected** by this bug today.
- If the destination is **blocked**, the packet is correctly dropped either way, just logged with the wrong protocol number.

So: real breakage is confined to *direct-routed* (non-WireGuard) IPv6 flows that use any IPv6 extension header — they get silently blackholed today. This is on top of the separate, more serious memory-safety issue: the walk has no bounds checking at all against the packet buffer, so a header's declared length can walk `ext`/`payload` past the end of the buffer.

## What changed

Extracted the walk into `app/src/main/jni/netguard/ip6_ext.{h,c}` — pure C, no JNI/Android headers, following the `dns_frame.c`/`dns_frame.h` extraction precedent (PR #773's `HostsBlocklistLogic` pattern) — so it can be unit-tested on the host. `handle_ip()` (`ip.c`) now calls `ip6_skip_ext_headers(pkt, length, &protocol, &payload_off)` and uses `payload_off` directly (fixing the double-add of `sizeof(ip6_hdr)` that existed in the original bail-out arithmetic, latent because the loop never reached a state where it mattered).

Fixes applied:
- **Hdr Ext Len arithmetic**: RFC 8200's 8-octet units, not raw octets — `8 * (len + 1)`, not `8 + len`.
- **Hard bounds checks**: every read of a header's own 2-byte prefix is bounds-checked before it happens, and a header's declared length is checked against the remaining buffer before advancing — `off` never exceeds `length`.
- **Loop termination**: rewritten so `protocol`/`payload_off` only ever indicate success when an upper-layer protocol was actually found.
- **Cap on headers walked**: `MAX_IP6_EXT_HEADERS = 8` bounds a maliciously (or corruptly) long chain.
- **Clean stop states**: on any "no parseable L4" outcome, `protocol_out` is set to the header/type the walk stopped on (never an `IPPROTO_TCP/UDP/ICMP/ICMPV6` value), and `payload_off_out` is always `<= length` — so a caller that only special-cases those four values (as `handle_ip()` does) treats a stopped walk exactly like today's "unknown protocol" path, never misreading extension-header bytes as a transport header.

Header types walked vs. stopped, and why:
- **Walked** (RFC 8200 8-octet units): Hop-by-Hop (0), Routing (43), Destination Options (60).
- **Walked** (RFC 4302 4-octet units, `(len + 2) * 4`): Authentication Header (51) — AH only authenticates, it doesn't encrypt, so a transport-mode AH header still has plaintext headers behind it worth walking into.
- **Stopped, not walked**: Fragment (44) — its `Hdr Ext Len`-shaped byte is actually a reserved field, not a length, so it cannot be walked the same way, and a non-first fragment has no upper-layer header at all. ESP (50) — payload is encrypted, nothing after it is parseable in plaintext. No Next Header (59) — a clean, legitimate end of chain. Mobility (135) was previously listed as walkable; dropped (out of scope/rare, and not safely walkable the same way).
- Also fixed: the original `is_lower_layer()` listed Destination Options (60) twice and never actually needed to (`is_upper_layer` supersedes it structurally in the new design).

No new preferences or user-facing options — this is a pure correctness/safety fix inside the native packet engine, consistent with the project's "simplicity over configurability" bias.

## Testing

Added `app/src/test/native/ip6_ext_test.c` (mirrors the existing `dns_frame_test.c` host-test pattern) and a CI step in `.github/workflows/test.yml` right after the DNS-framing one. Covers: no extension headers (plain IPv6, the common case); a single Hop-by-Hop header then TCP; a chain of three headers (Hop-by-Hop → Routing → Destination Options → TCP); AH's distinct 4-octet-unit length, followed by UDP; a declared header length that runs past the buffer; a truncated packet (header start present, no room for its own length byte); Fragment; ESP; No Next Header; and a chain long enough to hit the 8-header cap (with a real TCP header planted right where an unbounded walk would wrongly reach it).

Built and verified locally:
- `cc -Wall -Wextra -Werror -Iapp/src/main/jni/netguard -o /tmp/ip6_ext_test app/src/test/native/ip6_ext_test.c app/src/main/jni/netguard/ip6_ext.c && /tmp/ip6_ext_test` → all 10 tests pass.
- `./gradlew :app:compileGithubDebugJavaWithJavac` → succeeds.
- `./gradlew assembleGithubDebug` → succeeds for all four ABIs (native C + Rust); produced `TrackerControl-githubDebug-latest.apk`.
- `./gradlew :app:testGithubDebugUnitTest` → 279 tests, 0 failures, 0 errors.

**Negative control**: re-implemented the *old* buggy inline walk (same wrong `8 + len` arithmetic and the same broken loop-termination logic) under the new function signature and linked it against the *same*, unmodified test file. Result: 10 of the new assertions fail against it —
```
FAIL: single hop-by-hop: succeeds
FAIL: single hop-by-hop: protocol is TCP
FAIL: single hop-by-hop: payload after the 8-byte header
FAIL: chain of three: succeeds
FAIL: chain of three: protocol is TCP
FAIL: chain of three: payload after all three headers
FAIL: AH then UDP: succeeds
FAIL: AH then UDP: protocol is UDP
FAIL: AH then UDP: payload after AH's 4-octet-unit length
FAIL: long chain: payload_off right before the unreached 9th header
ip6_ext_test: 10 assertion(s) failed
```
This confirms the tests genuinely exercise the loop-termination bug (defect 4) — even the single-extension-header and three-header-chain cases, which is the dominant real-world failure mode — not just passing vacuously against the new code.

## Out of scope

- Adding userspace forwarding for Fragment/ESP flows on the direct (non-WireGuard) path — the engine has never had `handle_fragment`/`handle_esp`, and building that is unrelated to fixing the walk itself.
- Reassembling fragmented IPv6 packets to find the upper-layer header on a non-first fragment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,6 +73,13 @@ jobs:
             app/src/test/native/dns_frame_test.c app/src/main/jni/netguard/dns_frame.c
           /tmp/dns_frame_test
 
+      - name: Run IPv6 extension header walk host tests
+        run: |
+          cc -Wall -Wextra -Werror -Iapp/src/main/jni/netguard \
+            -o /tmp/ip6_ext_test \
+            app/src/test/native/ip6_ext_test.c app/src/main/jni/netguard/ip6_ext.c
+          /tmp/ip6_ext_test
+
       - name: Run unit tests
         run: ./gradlew testFdroidDebugUnitTest --offline
 
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -12,6 +12,7 @@ add_library( netguard
              src/main/jni/netguard/netguard.c
              src/main/jni/netguard/session.c
              src/main/jni/netguard/ip.c
+             src/main/jni/netguard/ip6_ext.c
              src/main/jni/netguard/policy.c
              src/main/jni/netguard/tls.c
              src/main/jni/netguard/tcp.c
--- a/app/src/main/jni/netguard/ip.c
+++ b/app/src/main/jni/netguard/ip.c
@@ -19,6 +19,7 @@
 
 #include "netguard.h"
 #include "tls.h"
+#include "ip6_ext.h"
 #include <stdatomic.h>
 
 int max_tun_msg = 0;
@@ -267,27 +268,6 @@ int check_tun(const struct arguments *args,
     return 0;
 }
 
-// https://en.wikipedia.org/wiki/IPv6_packet#Extension_headers
-// http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
-int is_lower_layer(int protocol) {
-    // No next header = 59
-    return (protocol == 0 || // Hop-by-Hop Options
-            protocol == 60 || // Destination Options (before routing header)
-            protocol == 43 || // Routing
-            protocol == 44 || // Fragment
-            protocol == 51 || // Authentication Header (AH)
-            protocol == 50 || // Encapsulating Security Payload (ESP)
-            protocol == 60 || // Destination Options (before upper-layer header)
-            protocol == 135); // Mobility
-}
-
-int is_upper_layer(int protocol) {
-    return (protocol == IPPROTO_TCP ||
-            protocol == IPPROTO_UDP ||
-            protocol == IPPROTO_ICMP ||
-            protocol == IPPROTO_ICMPV6);
-}
-
 // SNI extraction disabled by default: connecting to tracker IPs to read TLS
 // ClientHello leaks the user's IP address to the tracker server.
 // Can be enabled at runtime via jni_sni() for research purposes.
@@ -363,31 +343,18 @@ void handle_ip(const struct arguments *args,
 
         struct ip6_hdr *ip6hdr = (struct ip6_hdr *) pkt;
 
-        // Skip extension headers
-        uint16_t off = 0;
-        protocol = ip6hdr->ip6_nxt;
-        if (!is_upper_layer(protocol)) {
-            log_android(ANDROID_LOG_WARN, "IP6 extension %d", protocol);
-            off = sizeof(struct ip6_hdr);
-            struct ip6_ext *ext = (struct ip6_ext *) (pkt + off);
-            while (is_lower_layer(ext->ip6e_nxt) && !is_upper_layer(protocol)) {
-                protocol = ext->ip6e_nxt;
-                log_android(ANDROID_LOG_WARN, "IP6 extension %d", protocol);
-
-                off += (8 + ext->ip6e_len);
-                ext = (struct ip6_ext *) (pkt + off);
-            }
-            if (!is_upper_layer(protocol)) {
-                off = 0;
-                protocol = ip6hdr->ip6_nxt;
-                log_android(ANDROID_LOG_WARN, "IP6 final extension %d", protocol);
-            }
-        }
+        // Skip extension headers. ip6_skip_ext_headers() (ip6_ext.c) owns the
+        // walk -- RFC 8200 Hdr Ext Len arithmetic, hard bounds checks against
+        // `length`, and deciding which header types are walkable -- so it can
+        // be unit-tested on the host; see ip6_ext.h for the exact contract.
+        size_t payload_off;
+        if (!ip6_skip_ext_headers(pkt, length, &protocol, &payload_off))
+            log_android(ANDROID_LOG_WARN, "IP6 extension %d not walkable", protocol);
 
         saddr = &ip6hdr->ip6_src;
         daddr = &ip6hdr->ip6_dst;
 
-        payload = (uint8_t *) (pkt + sizeof(struct ip6_hdr) + off);
+        payload = (uint8_t *) (pkt + payload_off);
 
         // TODO checksum
     } else {
--- a/app/src/main/jni/netguard/ip6_ext.c
+++ b/app/src/main/jni/netguard/ip6_ext.c
@@ -1,0 +1,136 @@
+/*
+    This file is part of NetGuard.
+
+    NetGuard is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    NetGuard is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with NetGuard.  If not, see <http://www.gnu.org/licenses/>.
+
+    Copyright 2015-2019 by Marcel Bokhorst (M66B)
+*/
+
+#include "ip6_ext.h"
+
+#include <netinet/in.h>
+
+/*
+ * Extension header "Next Header" values that are not IPPROTO_* upper-layer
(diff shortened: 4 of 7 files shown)


kasnder:
Pushed `010524a2` on top: two extra host tests, because the suite as written could not detect the defect this PR is mainly about.

Every walkable-chain case used `Hdr Ext Len 0`, and at 0 the RFC 8200 form `8 * (len + 1)` and the original `8 + len` both give 8. Reverting just that one line in `ip6_ext.c` and re-running left the whole suite green. (The negative control reported in the PR body replaced the *entire* old loop, so it failed for bounds and termination reasons that masked the arithmetic being uncovered.)

Added a Hop-by-Hop with `Hdr Ext Len 1` (16 bytes; buggy form gives 9) and a Routing header with `Hdr Ext Len 3` (32 bytes; buggy form gives 11). With only the arithmetic line reverted, exactly those two now fail:

```
FAIL: hop-by-hop with options: advance is 8 * (len + 1), not 8 + len
FAIL: routing with options: advance is 8 * (len + 1), not 8 + len
ip6_ext_test: 2 assertion(s) failed
```

Also verified independently, on top of what the PR body reports:

- `cc -Wall -Wextra -Werror` clean, and the suite passes under `-fsanitize=address`.
- Two further isolated negative controls: walking AH with 8-octet units instead of 4 fails exactly the AH assertion; disabling the `advance > length || off > length - advance` bounds check fails exactly the four declared-length-past-end assertions.
- `assembleGithubDebug` builds `libnetguard.so` for all four ABIs, and `is_lower_layer`/`is_upper_layer` have no remaining callers anywhere in `app/src/main/jni/`.
- The `payload_off == length` boundary is safe at the call site: the ICMP, UDP and TCP branches in `handle_ip()` each check `length - (payload - pkt)` against the header size before dereferencing.

Issue #701, opened 2026-08-03
Add support for Android 17 Local Network Protections


anivicivokki:
**Do you use TrackerControl from Google Play/F-Droid/other?**
F-Droid, App version 2026.07.27-fdroid

**Describe the bug**
On Android 17, after upgrading to the new version, TrackerControl loses the ability to reach local network (e.g., LAN-hosted DNS servers, Pi-hole, AdGuard Home), possibly because the app does not declare or request the new `ACCESS_LOCAL_NETWORK` runtime permission.

This is a regression that breaks existing functionality for users who:
- Self-host DNS servers (Pi-hole, AdGuard Home, Unbound, etc.)
- Use custom local DNS for privacy/network control
- Want to use encrypted DNS (DoH/DoT) pointed at a local resolver

**To Reproduce**
1. Go to 'Advanced options'
2. Define VPN DNS pointing to local address (e.g., 192.168.x.x, 10.x.x.x, 172.16.x.x)
3. Try opening any address on browser, or using any other app.

PR #798, opened 2026-08-28
Warn when tunnelled traffic to the LAN is being dropped

### LINKED ISSUE #785: Validate on Android 17: LAN/IPv6 resolvers are host-routed into the tun, and the warning is suppressed if that breaks
Needs validation on a real Android 17 device — filing so it isn't lost. Came out of triaging user reports after the `2026080501` release.

## What the code does

`getBuilder()` pulls site-local IPv4 and all IPv6 resolvers into the tunnel as host routes (`ServiceSinkhole.java:1747-1760`), so the network's own resolver is re-originated from TrackerControl's socket. Since `targetSdk` moved to 37 (`app/build.gradle:39,44`), Android 17 Local Network Protections apply: traffic the app itself sends to RFC1918 destinations needs `ACCESS_LOCAL_NETWORK`, otherwise TCP times out and UDP fails with `EPERM`.

The code banks on the AOSP exemption for port 53 to the network's own resolvers (`LocalNetworkAccess.java:62-66`), and `isConfigured()` deliberately ignores system DNS.

## The concern

If a hardened fork (GrapheneOS was the reported platform) narrows or omits that exemption, every app behind TrackerControl loses name resolution — and the app does not warn, because `isMissing()` returns false for exactly this case. Silent total DNS failure with no diagnostic.

Note `LocalNetworkAccess` is otherwise advisory only: it never touches the `VpnService.Builder`, routes, or DNS selection (`ServiceSinkhole.java:1768-1773`, `ActivityMain.java:585`), so a missing grant cannot break routing by itself. The exposure is specifically the host-routed resolver path.

The IPv6 side arrived with `730a5520` ("Route IPv6 DNS through the VPN"), i.e. it predates the LNP work and has been in since `2026072701`.

## What would settle it

On an Android 17 device with `ACCESS_LOCAL_NETWORK` **not** granted, and a LAN resolver (router or Pi-hole):

1. Confirm whether name resolution works for apps behind the tunnel.
2. If it does not, check whether `LocalNetworkAccess.isMissing()` reports false while DNS is dead — that is the diagnostic gap, and is worth fixing regardless of whether the exemption holds.
3. Repeat with an IPv6-only mobile connection, which takes the second-pass branch at `ServiceSinkhole.java:1455-1462`.

kasnder:
Closes #785.

On Android 17 the local-network failure that actually reaches users is invisible
to us. Measured on a Pixel (Android 17, GrapheneOS) with `ACCESS_LOCAL_NETWORK`
**not** granted, probing the LAN router at 192.168.178.1 from `adb shell` — an
ordinary uid inside the tun, not TrackerControl's own:

| probe | permission denied | permission granted |
|---|---|---|
| DNS resolution (port 53) | works | works |
| TCP 80 / 443 / 8080 | timeout | connects |
| ICMP | 100% loss | 0% loss |

Two things follow, and both contradict comments we had in the tree:

1. Packets that enter the tun are re-sent from **TrackerControl's** socket, so
   **our** permission decides LAN reachability for every app behind the tunnel.
2. This is not confined to configured LAN destinations. `vpnStart()` adds a host
   route for each DNS server, so with the router as DNS — the common home setup
   — the router's whole address is pulled into the tun. DNS keeps working
   (Android exempts port 53 to the network's own resolvers), while the router
   UI, the NAS and the printer go dark for every app.

`isMissing()` could never see that: `isConfigured()` deliberately ignores the
network's own resolvers, and `observedLocalDestination` was only ever set from
our *own* outbound paths (`DnsOverHttpsClient`, `WgEgress`). So the user got no
banner and no notification — a silent, total loss of LAN access.

**Change:** report the destination of forwarded flows from `isAddressAllowed()`,
the per-flow native callback, so the existing banner and notification fire for
this case too.

The per-flow cost is a volatile read of a cached "permission is missing" flag
plus the existing latch check; the address is only parsed while both are true,
which is at most once. The permission itself is sampled in `isMissing()` — on
tunnel rebuilds and UI checks — never per packet.

Also corrects the two comments stating the refuted premise (the class javadoc of
`LocalNetworkAccess` and the #701 block in `vpnStart()`), and notes the known
gap that a LAN host reached over a *global* IPv6 address is not recognised as
local.

Tests: `LocalNetworkAccessTest` covers the latch, the non-local case, the reset,
and — via a value-taking `refreshPermission` seam, since Robolectric caps at API
36, below enforcement — that the per-flow gate is actually satisfiable for a
plain install with nothing configured.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -1767,15 +1767,15 @@ public class ServiceSinkhole extends VpnService {
                     Log.e(TAG, "addRoute DNS " + dns + ": " + ex);
                 }
 
-        // Android 17 refuses local network traffic without ACCESS_LOCAL_NETWORK:
-        // TCP times out, UDP fails with EPERM. A LAN resolver routed into the tun
-        // above is re-sent from our own socket, so it goes silent (#701).
-        // The symptom is "nothing resolves", which no one attributes to a
-        // permission — and the banner only reaches someone who opens the app.
+        // Android 17 applies local network protection to non-DNS traffic that
+        // enters the tun: TCP times out and UDP fails with EPERM for every app.
+        // DNS queries to port 53 survive the exemption, so a LAN resolver can
+        // keep resolving while other traffic silently fails (#701).
         // Notify, so the reason meets the user where the failure appears.
         if (net.kollnig.missioncontrol.LocalNetworkAccess.isMissing(ServiceSinkhole.this)) {
-            Log.w(TAG, "Local network access not granted: configured LAN destinations" +
-                    " (custom DNS, Secure DNS resolver, WireGuard peer, proxy) are unreachable");
+            Log.w(TAG, "Local network access not granted: LAN destinations are" +
+                    " unreachable, whether configured (custom DNS, Secure DNS resolver," +
+                    " WireGuard peer, proxy) or reached through the tun by another app");
             showLocalNetworkNotification();
         } else
             clearLocalNetworkNotification();
@@ -2591,6 +2591,12 @@ public class ServiceSinkhole extends VpnService {
             Log.i(TAG, "Blocking DoT " + packet);
             packet.allowed = false;
         }
+
+        // Report only traffic that will be forwarded, and never TC's own
+        // sockets. The helper's cheap gates run before it parses packet.daddr.
+        if (packet.allowed && packet.uid != Process.myUid()
+                && net.kollnig.missioncontrol.LocalNetworkAccess.shouldReportForwardedDestination())
+            net.kollnig.missioncontrol.LocalNetworkAccess.reportForwardedDestination(packet.daddr);
 
         if (packet.allowed)
             if (mapForward.containsKey(packet.dport)) {
--- a/app/src/main/java/net/kollnig/missioncontrol/LocalNetworkAccess.java
+++ b/app/src/main/java/net/kollnig/missioncontrol/LocalNetworkAccess.java
@@ -42,16 +42,19 @@ import java.net.URI;
  * permission for apps targeting API 37 or higher. Without it, TCP connections
  * time out and UDP fails with {@code EPERM}.
  *
- * <p>Traffic other apps send to the LAN is unaffected by TrackerControl's
- * permission state: those are their own sockets, and TrackerControl keeps RFC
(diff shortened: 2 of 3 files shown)


kasnder:
Opened #799 as an alternative that covers the motivating case without touching the packet path — worth comparing before this lands.

The device measurement here has been confirmed: the router at 192.168.178.1 *was* the network's DNS server, so `getBuilder()` gave it a `/32` host route. That makes the failure fully determinable at tunnel-build time. The loop at `ServiceSinkhole.java:1829` already iterates exactly the addresses that receive a route and already tests `dns.isSiteLocalAddress()`, so one `reportRoutedResolver(dns)` call inside it latches the observation before the `isMissing()` check a few lines below — no per-flow work at all.

Two things about the per-flow gate as written:

- **`shouldReportForwardedDestination()` is not "at most once".** It is `!observedLocalDestination && missingPermission`, and the latch only closes when a *local* address is seen. In the common case — Android 17, permission not granted, nothing local ever in the tun — it never latches, so `packet.daddr` is parsed on every forwarded flow for the life of the tunnel. That is cheap next to what `isAddressAllowed()` already does (a `getDefaultSharedPreferences()`, a `ReadWriteLock` acquire, `blockKnownTracker()`, `InternetBlocklist`, possibly a `logPacket()` insert) and it adds no wakeup or timer, so the battery cost is negligible either way — but the description's justification does not match the code.
- **"the router UI, the NAS and the printer go dark" is broader than the mechanism supports.** The host route covers the resolver's address only, and `VpnRoutes` still excludes the RFC 1918 ranges, so a NAS or printer on a *different* LAN address never enters the tun. Only services sharing the resolver's address are affected (which for the home-router case is still the web UI, so the user-visible bug stands).

What #799 does *not* cover, and this PR's hook would: LAN destinations pulled in by tethering compatibility mode's full-tunnel route. If that case is worth the per-flow gate, it seems better argued on its own terms — and the gate could then be scoped to it, rather than running whenever the permission is missing. The remaining case in the description, a LAN host on a global IPv6 address, is not detected by either approach, since `isLocalAddress()` cannot recognise it.

The comment corrections and the `isLocalAddress()` javadoc note here are good regardless; #799 carries equivalents so the two do not depend on each other.


---

kasnder:
Superseded by #799, which covers the routed-resolver failure at tunnel-build time without adding work to the forwarded-flow path and keeps resolver state scoped to the current tunnel.

PR #803, opened 2026-08-28
Retry failed WireGuard startup

kasnder:
## Summary
- retry failed WireGuard startup with a separate five-attempt exponential-backoff budget
- keep captured traffic fail-closed while startup is unavailable
- cancel/reset retries across success, stop, revoke, destroy, inactive users, and stale commands
- use uptime consistently so deep sleep cannot reset the budget

## Why
A transient `WgEgress.startOrUpdate` failure left the VPN descriptor active and traffic blocked, but never attempted automatic recovery.

## Validation
- focused startup, native recovery, VPN replacement, and sequencing JVM tests
- Java compile
- `git diff --check`

## Risk
- No device lifecycle/wake-lock test was run. Retries are bounded and use no polling loop or forced keepalive; the existing error remains visible after exhaustion.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -260,6 +260,49 @@ public class ServiceSinkhole extends VpnService {
         ServiceSinkhole.reload("native tunnel recovery", ServiceSinkhole.this, false);
     };
 
+    private static final int WG_STARTUP_RECOVERY_MAX_RETRIES = 5;
+    private static final long WG_STARTUP_RECOVERY_INITIAL_DELAY_MS = 1_000L;
+    private static final long WG_STARTUP_RECOVERY_STABLE_WINDOW_MS = 2 * 60_000L;
+    private static final String EXTRA_WG_STARTUP_RETRY = "WireGuardStartupRetry";
+    private final WireGuardStartupRecoveryPolicy wgStartupRecoveryPolicy =
+            new WireGuardStartupRecoveryPolicy(
+                    WG_STARTUP_RECOVERY_MAX_RETRIES,
+                    WG_STARTUP_RECOVERY_INITIAL_DELAY_MS,
+                    WG_STARTUP_RECOVERY_STABLE_WINDOW_MS);
+    private final Object wgStartupRecoveryLock = new Object();
+    private final Handler wgStartupRecoveryHandler = new Handler(Looper.getMainLooper());
+    private final Runnable wgStartupRecoveryRunnable = () -> {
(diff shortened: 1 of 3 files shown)


kasnder (commented):
## Review

**Verdict: approve with minor nits.** Reviewed in a fresh worktree at `87315b2a`.

### Verified working

- **Retry path is real**: with the descriptor active and an unchanged builder, the retry `reload()` takes the "Native restart" branch and `startNative` re-attempts `WgEgress.startOrUpdate` (`ServiceSinkhole.java:865`, `:2145`).
- **Two-phase `claim()`/`begin()` marker is sound.** Without `begin()`, a dispatched-but-dropped command would strand `dispatched=true` and block all later `onFailure` scheduling. Every early return after `begin()` (background user, temporary stop, switch cases, both catch blocks, `stop()`) cancels correctly; stale retries are dropped by `begin()` returning false; untagged reloads clear pending via `cancel(false)` without eating the budget.
- **Lock ordering is consistent** (`ServiceSinkhole.this` → `wgStartupRecoveryLock`, never reversed) — no deadlock. Wakelock + `commandHandler.queue` matches the VPN-replacement runnable; main-looper handler matches the existing `nativeRecoveryHandler` pattern.
- **Uptime clock is correct**: `Handler.postDelayed` is uptime-based, so policy and dispatch agree; deep sleep can't age an episode into the stable window.
- **Fail-closed preserved**: `startNative` returns false before `jni_start`, the tun stays up, traffic stays blocked.

### Checks run

- `:app:compileGithubDebugJavaWithJavac` + `WireGuardStartupRecoveryPolicyTest` — pass.
- `git diff master...pr-803 --check` — clean; mergeable/clean vs master.

### Nits (non-blocking)

1. `WireGuardStartupRecoveryPolicy.onSuccess()` is dead code — the success path uses `cancelWireGuardStartupRecovery(true)` (`ServiceSinkhole.java:2161`), which is functionally identical. Call `onSuccess()` there or delete the method; the test `successfulStartupResetsBackoff` currently exercises only dead code.
2. Any exception in a retry-tagged command *before* `startNative` (e.g. `Rule.getRules`) resets the budget via the catch handlers, so a later failure gets a fresh 5-retry budget. Bounded and acceptable, but worth knowing.
3. Pre-existing, optional follow-up: `scheduleVpnReplacementRecovery` uses `elapsedRealtime()` with uptime-based `postDelayed` — exactly the mismatch this PR's comment fixes for WG.

### Non-blockers

- Exhaustion only logs (no `msg_vpn_replacement_recovery_exhausted`-style notification like VPN replacement has) — acceptable since the persistent WG error notification already covers the user.
- No device lifecycle/wake-lock test (acknowledged in the PR description).

PR #804, opened 2026-08-28
Publish tracker cache entries atomically

kasnder:
## Summary
- replace the parallel IP-to-host and IP-to-tracker maps with one immutable cache entry
- make invalidation atomic with generation-checked publication
- preserve the exact cache snapshot through tracker decisions and log output
- clear verdicts after tracker-data reloads

## Why
The old two-map cache could expose a hostname without its tracker verdict (or vice versa), and an in-flight database read could republish stale attribution after invalidation.

## Validation
- `./gradlew :app:testGithubDebugUnitTest -q`
- focused `TrackerCacheTest` race, expiry, null, and invalidation coverage
- `git diff --check`

## Risk
- No device packet-path profiling was performed. Live hits remain a single `ConcurrentHashMap.get`; mutation serialization occurs only on misses and invalidation.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -126,7 +126,6 @@ import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.zip.GZIPInputStream;
 
@@ -906,10 +905,7 @@ public class ServiceSinkhole extends VpnService {
             DatabaseHelper.getInstance(ServiceSinkhole.this).cleanupDns();
 
             // Refresh mappings regularly
-            ipToHost.clear();
-            ipToTracker.clear();
-            // Same race as dnsResolved(): invalidate after clearing.
-            trackerCacheGeneration.incrementAndGet();
+            trackerCache.clear();
             uidToApp.clear();
             uidToPackage.clear();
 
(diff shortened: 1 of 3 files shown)


kasnder:
Reviewed the diff and the surrounding packet-path code; verified the concurrency claims against the actual call sites.

**What this fixes (all real issues):**
- **Half-publication:** `ipToHost` and `ipToTracker` could be observed independently, so a reader could hit host-cache-hit/tracker-cache-miss and silently return "no verdict" (skipping a block) instead of re-reading the DB. One immutable entry removes that state entirely.
- **Residual TOCTOU:** the old check-generation-then-two-puts pattern could land stale entries *after* a concurrent `clear()`+bump — the generation check had already passed and nobody re-checked. Serializing `putIfGeneration` against `invalidate`/`clear` under one lock closes it, and `get()` stays lock-free on the packet path.
- **Cache-clear gap:** `prepareHostsBlocked` reloaded the TrackerList without clearing the caches, while the `ActivitySettings` and `BlockingMode.applyMode` call sites already did. The new `clearTrackerCaches()` fills that gap.

**Verified correct:**
- `Entry`'s non-null invariants hold on the only construction path: `dname` falls back to `NO_DNAME` and `tracker` to `NO_TRACKER` before `new Entry(...)` (ServiceSinkhole.java:2908-2911 in the PR), so the `requireNonNull` checks can't NPE the packet path.
- Expired-entry removal uses two-arg `remove(address, entry)`, so a stale expiry sweep can't evict a newer entry; the expiry boundary (`now > expires`) matches the old `Expiring` semantics; pure-refresh rows still skip invalidation (#802 behavior preserved).
- No remaining references to the removed `ipToHost`/`ipToTracker`/`Expiring`.

**Nits (non-blocking):**
- The `tracker != null` guard at the logging/blocking branch is now dead code (tracker can't be null there anymore) — harmless if kept as defense.
- Ignoring `putIfGeneration`'s return value is intentional and commented — fine.
- Cached `Tracker` instances are the shared, mutable registry instances (public fields); pre-existing, not worsened here.

**Tests:** full `:app:testGithubDebugUnitTest` — 43 suites, 0 failures, including the 10 new `TrackerCacheTest` tests.

PR #805, opened 2026-08-28
Publish tracker-list reloads atomically

kasnder:
## Summary
- build tracker assets in a private map and publish map, mode flags, and effective blocking mode as one snapshot
- retain the previous complete snapshot when an asset load fails
- make singleton initialization safe and avoid double-loading assets
- clear service verdict caches only after successful publication

## Why
The previous reload cleared the live tracker map before rebuilding it, so packet-path readers could observe partial data and cache a false non-tracker verdict for a DNS TTL.

## Validation
- focused data, cache, DNS-attribution, and blocking-mode JVM tests
- concurrent singleton and failed-reload mode-retention coverage
- `git diff --check`

## Dependency
- Stacked on #804 because atomic reload publication also requires epoch-safe verdict-cache invalidation.

## Risk
- The hosts-file map itself remains independently managed; this PR makes tracker-list publication atomic and does not redesign hosts-file loading.

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
+++ b/app/src/main/java/eu/faircode/netguard/ServiceSinkhole.java
@@ -2931,7 +2931,7 @@ public class ServiceSinkhole extends VpnService {
         }
 
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
-        String blockingMode = BlockingMode.getMode(this);
+        String blockingMode = TrackerList.getBlockingMode(this);
         boolean blockAmbiguousTrackers = BlockingModeLogic.blocksAmbiguousTrackerIp(blockingMode);
         Tracker tracker = null;
         TrackerCache.Entry cached = trackerCache.get(daddr, System.currentTimeMillis());
--- a/app/src/main/java/net/kollnig/missioncontrol/data/BlockingMode.java
+++ b/app/src/main/java/net/kollnig/missioncontrol/data/BlockingMode.java
@@ -247,8 +247,12 @@ public class BlockingMode {
         if (trackerBlocklist.applyStrictModeToAll(isStrictMode(c)))
             trackerBlocklist.saveSettings(c);
 
-        ServiceSinkhole.clearTrackerCaches();
-        TrackerList.reloadTrackerData(c);
+        if (TrackerList.reloadTrackerData(c)) {
+            // Clear caches only after the new tracker snapshot is published. A
+            // pre-reload clear allows packet-path lookups during asset loading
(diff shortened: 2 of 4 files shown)


kasnder (commented):
**Verdict: approve** — correct fix for a real race, reviewed from a local build (full `:app:testGithubDebugUnitTest` passes).

The old `reloadTrackerData` cleared the live map before rebuilding it, so packet-path `findTracker` calls during a reload could see an empty/partial map and cache a false non-tracker verdict for up to a full DNS TTL (7 days). Snapshot-swap publication fixes that; failure now retains the previous complete snapshot (previously a failed load silently left an empty map). The reader path stays lock-free (one volatile read) — no new locking on the packet path. The `putIfAbsent` in `findTracker` also closes a latent duplicate-`Tracker` race in hostlist caching. Clearing verdict caches *after* publication is the right order, and the generation check from #804 is what makes it safe against a pre-publication lookup reinserting a stale verdict. Reading the mode from the same published snapshot (`TrackerList.getBlockingMode`) keeps blocking semantics in sync with the loaded lists.

Non-blocking notes:

1. **Silent mode divergence on failed reload** (`BlockingMode.applyMode`): prefs now hold the new mode while the retained snapshot keeps the old one, so UI and effective blocking disagree until the next successful reload (hosts update, next mode change, app restart) — only a `Log.w` today. Self-healing and fail-safe (old map ↔ old semantics stay consistent), and the test codifies the retention as intended, but a user-visible signal or a retry on the next `ServiceSinkhole.reload` would be nicer if this ever fires in practice.
2. **Residual one-packet skew**: `blockKnownTracker` reads `blockingMode` from the snapshot at entry while `findTracker` re-reads the latest snapshot per call — a reload publishing mid-decision can pair the old mode with the new map for one packet. Self-healing on the next packet; threading the snapshot through `findTracker` isn't worth the churn.
3. `getBlockingMode` javadoc slightly overstates ("never read from two different lifecycle states") — see note 2.
4. `ActivitySettings.java:984` ignores the new boolean and clears unconditionally; harmless (the snapshot only changes on success), but honoring it would be more consistent. Same for `ServiceSinkhole.java:2342`, though there unconditional clearing is correct since the hosts file changed regardless.
5. The failure contract covers `IOException | IllegalStateException` only; other RuntimeExceptions (e.g. from `getAssets().open`) still propagate as before the PR — fine, but the new `boolean` return may suggest broader coverage to future readers.
6. Pre-existing, not this PR: duplicate `import eu.faircode.netguard.DatabaseHelper` in `TrackerList.java`.

Process: stacked on #804 — needs merge in that order; note this PR currently shows no CI checks, worth triggering before merge.

PR #806, opened 2026-08-28
Recover WireGuard monitoring after stats failures

kasnder:
## Summary
- retry transient initial and mid-loop WireGuard stats failures at the existing screen-aware cadence
- make monitor stop/restart and callback ownership generation-safe
- detect a stats call that has stopped making uptime progress when liveness is queried
- restart a dead watchdog on the same-config tunnel path

## Why
A null initial stats read permanently ended monitoring, and lifecycle races could let an old monitor affect a replacement.

## Validation
- all `net.kollnig.missioncontrol.wg.*` JVM tests
- Kotlin and Java compile
- latch-controlled stop/stale/hung-stats coverage
- `git diff --check`

## Risk
- No device test simulated a native `Tunnel.stats()` hang. Detection adds no timer, alarm, radio work, or faster screen-off cadence.

### DIFF
--- a/app/src/main/java/net/kollnig/missioncontrol/wg/WgConnectivityMonitor.kt
+++ b/app/src/main/java/net/kollnig/missioncontrol/wg/WgConnectivityMonitor.kt
@@ -178,7 +178,7 @@ internal class WgConnectivityChecker(private val prod: () -> Unit) {
         if (!cadenceOk) return
         try {
             prod()
-        } catch (e: Throwable) {
+        } catch (e: Exception) {
             Log.w(TAG, "prod threw", e)
         }
         if (initialProdTs == null) initialProdTs = now
@@ -216,7 +216,7 @@ private const val TAG = "WgConnMonitor"
  */
 internal class WgConnectivityMonitor(
     private val statsProvider: () -> WgStats?,
-    prod: () -> Unit,
+    private val prod: () -> Unit,
     private val onBroken: () -> Unit,
     // Fired each time this monitor observes a completed handshake after a
     // period without one (tunnel start, or re-handshake after session expiry),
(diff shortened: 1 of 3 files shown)


kasnder (commented):
## Review

**Verdict: approve with one design question.**

### Solid

- The core zombie bug is real and correctly fixed: master's `statsProvider() ?: return` exits the run loop leaving `running=true`, so the tunnel stays up but is permanently unwatched. The bounded retry + `startMonitorIfDead` close it.
- Per-run `WgConnectivityChecker`, generation-gated callbacks, the epoch-based `WgMonitorLifecycle`, and the `stopInternal` reorder (invalidate tunnel before stopping the monitor) all check out. Traced the lock ordering (`callbackLock` → `lifecycleLock`; `WgMonitorLifecycle` never holds `monitorLock` across `stop()`) — no inversion or deadlock found, including reentrant stop-from-callback.
- Native premise confirmed: `Tunnel.stats()` → `Err("tunnel stopped")` on a teardown race (`wgbridge-rs/src/tunnel.rs`), the JNI layer throws, `statsOrNull` swallows → null. Transient null stats are real.

### Finding 1 — needs a decision: BROKEN-at-idle deferral

- The checker-BROKEN path now defers recovery while the screen is off (`WgConnectivityMonitor.kt:549-553`). On master an overnight dead tunnel was rebind/restarted within ~30 s; now it is only prodded every ~20 s until screen-on, indefinitely.
- This contradicts the `IDLE_LOOP_SLEEP_MS` doc comment ("a background-sync stall is caught within one cycle and recovery still runs", line 254), has no test, and isn't mentioned in the PR body.
- Battery-wise the deferral is defensible (no wakelock/DNS/handshake churn at night) — but it should be an explicit, commented decision (and that comment updated), not a side effect of the stats-retry work.

### Finding 2 — minor: new stop latency

- `stop()` now blocks on `callbackLock`, so VPN teardown (`stopInternal` → `stopMonitor`) can wait out one in-flight callback — worst case a cheap recovery doing bounded DNS (4 s/peer). Bounded and short in practice since callbacks re-check `isCurrent`, but it is new teardown latency worth being aware of.

### Nits

- The two spin-waits use `return@repeat`, which continues rather than breaks (`WgConnectivityMonitorTest.kt:328`, `:377`) — the assertions are still valid, but the early-exit intent isn't.
- `WgVerdict.GONE -> return` is unreachable in the loop (stats non-null is guaranteed before `tick`) — fine as defense.
- Throwable→Exception narrowing means an Error kills the monitor thread silently until the next `startMonitorIfDead` — acceptable, and consistently applied.

### Verification

- Full `net.kollnig.missioncontrol.wg.*` JVM suite passes locally; `WgConnectivityMonitorTest` survived 3 `--rerun-tasks` repeats with no flake. CI (test + instrumentation) green.

PR #807, opened 2026-08-29
Fix direct DNS-over-TCP stream framing

kasnder:
## Summary
- reassemble DNS-over-TCP as a byte stream across split and coalesced socket reads
- apply blocking rewrites to every complete frame before any part reaches the app
- bound reassembly and pending output while respecting the TCP window, MSS, and sequence accounting
- preserve queued output across ACK, HUP, EOF, and error transitions without periodic polling

## Why
`recv()` boundaries are not DNS frame boundaries. The previous path could miss later coalesced answers or parse partial answers, causing tracker detection and blocking gaps.

## Validation
- host C suite with `-Wall -Wextra -Werror -pedantic`
- ASan/UBSan host run
- `:app:externalNativeBuildGithubDebug` for all configured ABIs
- Rust workspace and `tc-dns` C-ABI tests
- `git diff --check`

## Risk
- No device-level DNS-over-TCP capture was run. Allocation failure forwards unchanged when the exact stream can be queued; if pending storage itself is unavailable, the TCP session closes instead of forwarding a partial/corrupt stream.

### DIFF
--- a/app/src/main/jni/netguard/dns_frame.c
+++ b/app/src/main/jni/netguard/dns_frame.c
@@ -19,39 +19,333 @@
 
 #include "dns_frame.h"
 
-struct dns_frame_decision dns_frame_decide(size_t bytes, size_t frame_len) {
-    struct dns_frame_decision d;
-    d.frame_len = frame_len;
+#include <string.h>
 
-    if (frame_len == 0) {
-        // A zero-length declared frame is not something the parser can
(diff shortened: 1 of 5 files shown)


kasnder:
Too long, won't merge. DNS over TCP is relatively rare.

kasnder:
Closing this oversized prototype for now. Follow-up work is tracked in #809, with this PR retained as design, implementation, and test evidence for a smaller reviewable series.

PR #819, opened 2026-08-29
Warn immediately when allowed DoT is observed

kasnder:
### Motivation

Encrypted DNS can silently bypass TrackerControl's DNS-based detection when DoT is used while DoT blocking is disabled. Users should be warned as soon as evidence of that bypass appears, without adding periodic work.

### Description

- Add evidence-ordered Private DNS bypass warning states for hostname mode and observed allowed DoT.
- Show the warning in the main-screen banner and a transition-based notification.
- Gate observed-DoT detection on the full traffic log and recent allowed port-853 evidence.
- Re-evaluate the warning immediately when an allowed DoT flow is logged, rather than waiting for VPN establishment or a network change.
- Add Robolectric coverage for mode combinations, precedence, log gating, recency, and transitions.

### Testing

- `git diff --check` (passed)
- `./gradlew :app:testGithubDebugUnitTest --tests 'eu.faircode.netguard.PrivateDnsWarningTest' -q` (environment limitation: Android SDK is not configured)
- `./gradlew :app:compileGithubDebugJavaWithJavac -q` (environment limitation: Android SDK is not configured)

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityMain.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityMain.java
@@ -492,6 +492,24 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
             }
         });
 
+        TextView tvPrivateDnsBypass = findViewById(R.id.tvPrivateDnsBypass);
+        tvPrivateDnsBypass.setVisibility(View.GONE);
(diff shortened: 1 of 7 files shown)


kasnder:
Closing this duplicate: the fix was pushed directly to the original PR #817.

PR #820, opened 2026-08-29
Warn immediately when allowed DoT is observed

kasnder:
### Motivation
- Encrypted DNS can silently bypass TrackerControl's DNS-based detection when DoT is used and DoT blocking is disabled, creating a false sense of security that should be surfaced to users as evidence appears. 
- The UI and notification surfaces should reflect two evidence-ordered cases (hostname-mode loud failure and allowed-DoT observed) without adding periodic work or changing existing precedence rules.

### Description
- Add evidence-driven Private DNS bypass handling to the service: new warning states (`HOSTNAME`, `ALLOWED_DOT`), notification/banner lifecycle, and helpers to show/clear/reset the bypass notification when network or preference state changes. 
- Add `DatabaseHelper.hasRecentAllowedDot(long)` to query the traffic log for recent allowed DoT (port 853) flows and gate the allowed-DoT warning on the full traffic-log preference. 
- Wire a UI banner row (`tvPrivateDnsBypass`) into `ActivityMain` and add the associated layout and string resources, plus `Util.isPrivateDnsDetectionDefeated()` to centralise the hostname-mode detection logic. 
- Add Robolectric unit tests (`PrivateDnsWarningTest`) that cover the Private DNS × `block_dot` matrix, precedence of hostname-mode over allowed-DoT, log gating, recency checks, and state-transition behaviour; and make a small fix so the service re-evaluates the bypass warning immediately when an allowed DoT flow is inserted into the traffic log.

### Testing
- `git diff --check` was run and produced no whitespace errors. (passed)
- `./gradlew :app:testGithubDebugUnitTest --tests 'eu.faircode.netguard.PrivateDnsWarningTest'` was attempted but the Gradle run failed because the Android SDK is not configured in this environment (missing `ANDROID_HOME`/`local.properties`), so unit tests could not be executed here. (not executed)
- `./gradlew :app:compileGithubDebugJavaWithJavac` was attempted but also failed for the same missing Android SDK configuration, so a local compile check could not be completed. (not executed)

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a92add04970832bb3f9ef64f6679531)

### DIFF
--- a/app/src/main/java/eu/faircode/netguard/ActivityMain.java
+++ b/app/src/main/java/eu/faircode/netguard/ActivityMain.java
@@ -492,6 +492,24 @@ public class ActivityMain extends AppCompatActivity implements SharedPreferences
             }
         });
 
+        TextView tvPrivateDnsBypass = findViewById(R.id.tvPrivateDnsBypass);
+        tvPrivateDnsBypass.setVisibility(View.GONE);
+        tvPrivateDnsBypass.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                Intent settings;
+                if (ServiceSinkhole.getPrivateDnsWarningState(ActivityMain.this)
+                        == ServiceSinkhole.PRIVATE_DNS_WARNING_ALLOWED_DOT)
+                    settings = new Intent(ActivityMain.this, ActivitySettings.class);
+                else {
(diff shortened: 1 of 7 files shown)


kasnder:
Closing this duplicate: the fix was pushed directly to the original PR #817.

PR #847, opened 2026-08-29
Redesign TRACKERS tab as calm category feed

### LINKED ISSUE #832: Company-first trackers view (design)
## Motivation

TrackerControl's per-app trackers list is grouped by **category** (Advertising,
Analytics, Content, …). Categories are the *control unit* — `TrackerBlocklist`
keys are `"<Category> | <Company>"`, and a bare category key means "allow the
whole category" — but they are not what users actually reason about. Users think
in companies ("is Google getting my data from this app?"), and today a single
company shows up once per category it was observed in, with no single place to
allow or block it.

The proposal is a **company-first view**: one row per company, with its observed
categories shown as metadata and the per-category detail nested inside.

## Why this is filed rather than implemented

Categories were originally introduced *because* long per-app company lists got
messy. A company-first list re-introduces exactly that problem, and it is worse
now: with `domain_based_blocking`, DuckDuckGo pseudo-companies multiply rows
further (one row per pseudo-company).

**This issue must not be implemented until it has an answer to scalability** —
e.g. sorting by recency, collapsing older entries behind a "show more", a search
field, or some combination. That is the open question, and it is the reason this
was postponed rather than built.

Everything below is a design that was reviewed against the codebase (file:line
anchors were checked at the time of writing, on top of the branch behind
https://github.com/TrackerControl/tracker-control-android/pull/826). It is
recorded so the reasoning is not lost, not because it is ready to build.

---

### 1. Pure logic — new `data/CompanyGrouping.java`

```java
CompanyRow {
    String company;
    List<Tracker> trackers;    // one per observed (category|company) pair
    List<String> categories;
    long lastSeen;             // max over pairs
    boolean uncertain;         // any pair uncertain
    boolean contentOnly;
    boolean allAmbiguousAllowed;
}
WritePlan { Set<String> keysToAllow; Set<String> keysToDisallow; }
```

`WritePlan` is applied through the existing `TrackerBlocklist.unblock(uid, key)` /
`block(uid, key)`. **Zero changes to `TrackerBlocklist`**; the key format is
untouched by construction.

- `group(List<TrackerCategory>)` — merge `getAppTrackers` output across
  categories, sorted by `lastSeen` descending.
- `isRowBlocked(row, allowedKeys, strict)` — a row is blocked iff every non-Content
  pair has *both* its category key and its company key absent from the allow set
  (a null set means everything is blocked). Content pairs participate only in
  strict mode.
- `allowCompany(row, strict)` — add the company key for every observed non-Content
  category (all categories in strict).
- `blockCompany(row, observed, allowedKeys, strict)` — the write-through: for each
  of the row's categories that carries a *bare category-allow* key, add
  company-allow keys for every **other** company observed for this app in that
  category, drop the bare category key, then remove this row's company keys.
  Content is untouched outside strict. Unobserved/future companies revert to
  blocked automatically, since expansion only ever adds keys for observed
  companies.

### 2. Rendering — `details/TrackersListAdapter.java`, `res/layout/list_item_trackers.xml`

One adapter, branching on a `group_by_company` preference — no parallel adapter
class. `set()` additionally computes `mCompanyRows = CompanyGrouping.group(items)`
when grouping is on.

- Company row bind: `root_name` = company name, substituting
  `R.string.tracker_hostlist` for `TRACKER_HOSTLIST` exactly as `updateText` already
  does (`TrackersListAdapter.java:321-324`). A new small `tvRowCategories` TextView
  (GONE in category mode) shows the joined category display names — this needs a
  static `TrackerCategory.getDisplayName(Context, String)` overload, since `LABELS`
  is private (`TrackerCategory.java:40`).
- Switch: enabled iff `currentState(w) == PROTECTED && !row.contentOnly` (outside
  strict) `&& !row.allAmbiguousAllowed`; checked = `isRowBlocked(...)`; the listener
  applies the `WritePlan` then `notifyDataSetChanged()` (the write-through
  normalizes state). The old "category unblocked" refusal toast does not apply in
  this mode. `contentOnly` rows outside strict: switch hidden, status
  "Allowed — essential services".
- Inner list = the row's per-category pairs (category name + per-pair
  Blocked/Allowed via the existing `blocked()` calls + hosts). No item click
  listener in company mode. Global minimal mode reuses the existing read-only
  branch. `domain_based_blocking` pseudo-companies get one row each (accepted;
  worth a comment in the code).
- Persistence unchanged — `DetailsActivity.savePrefs` in `onPause`.

### 3. Preference, migration, upgrade hint

- `res/xml/preferences.xml`, `category_options`:
  `SwitchPreference key="group_by_company"`, title `setting_group_by_company`,
  summary `summary_group_by_company`, `defaultValue=false`.
- `ApplicationEx.migratePreferences` (`ApplicationEx.java:201`): if the key is
  absent, `putBoolean("group_by_company", resolveGroupByCompanyDefault(prefs.getInt("version", -1)))`
  with a pure `resolveGroupByCompanyDefault(v) { return v < 0; }` — i.e. **new
  installs get the company view, upgrades keep categories**. (Fresh install ⇔
  `version == -1`, an invariant maintained by `ReceiverAutostart.upgrade`;
  precedent: `resolveBlockingModeMigration`.) Fresh installs also pre-dismiss
  `hint_group_by_company`.
- Upgrade snackbar in `TrackersFragment.onResume`: if `!group_by_company &&
  hint_group_by_company`, show `group_by_company_hint` with a "Settings" action
  opening `ActivitySettings`, and persist `hint_group_by_company = false`
  immediately (shown once).
- Export/import: `group_by_company` lives in default prefs and is exported
  wholesale already — nothing to add.
- New strings: `setting_group_by_company`, `summary_group_by_company`,
  `tracker_company_content_only` ("Allowed — essential services"),
  `group_by_company_hint`, `group_by_company_hint_action`,
  `toggle_block_company_description`.

### Tests the design assumed

`CompanyGroupingTest`: cross-category merge; `contentOnly`; `lastSeen`/`uncertain`
aggregation; `isRowBlocked` with a null subset / category-allow / company-allow /
strict-Content; `allowCompany` keys; `blockCompany` write-through including never
emitting a bare Content key outside strict, never adding unobserved companies,
idempotence, and the mixed-state row (allowed in one category shows as allowed, and
the first block normalizes it). A migration test as a sibling of
`BlockingModeMigrationTest` (fresh ⇒ true + hint pre-dismissed; upgrade ⇒ false;
existing key untouched). Guard re-runs of `TrackerBlocklistTest` (especially
`blockedTrackerRequiresBothCategoryAndKeyBlocked`) and `BlockingBaselineSubsetTest`.

## Known limitations of the design

- Blocking a company whose pairs are all ambiguous shared-IP has no runtime effect
  in standard mode (`ServiceSinkhole` allows mixed-evidence IPs before granular
  rules apply) — hence `allAmbiguousAllowed` disabling such rows, mirroring how
  individual rows are already disabled today.
- In strict mode a user's Content-company edits are overridden by the next mode
  switch to standard/minimal (`applyStrictModeToAll` re-adds the bare Content key).
  Consistent with "Content is mode territory"; no mitigation proposed.
- Rows with mixed pre-existing state render as allowed until the first block toggle
  normalizes them.

kasnder:
## Summary

- Replace nested per-category cards and eager list measurement with a flat, recycled category feed ordered by recent contact.
- Keep the full existing header, including browser warning, Protection routing summary, Libraries analysis row, and tap targets; name the active blocking level in the Protection row.
- Preserve TrackerBlocklist category and category-company keys without migration while adding read-only Minimal and MINIMAL_ONLY states, Monitored and shared-IP handling, expandable host details, and per-section collapse after 10 companies.
- Add pure feed/status logic with branch and collapse edge-case tests; remove the obsolete nested-list layouts.

## Verification

- :app:testGithubDebugUnitTest with focused TrackerFeedLogicTest and TrackerStatusLogicTest
- :app:compileGithubDebugJavaWithJavac
- :app:testGithubDebugUnitTest full suite
- :app:lintGithubDebug
- Fresh read-only gpt-5.6-sol adversarial review; verified findings corrected

## Remaining risk

- No device or emulator smoke test was run, so final interaction and visual checks across Standard, Strict, Minimal, paused, browser, Libraries, and WireGuard-route states remain for manual QA.

Closes #832


kasnder:
Reviewed against the plan agreed in #832 (Option E). **The PR implements it faithfully and builds clean** — `:app:compileGithubDebugJavaWithJavac`, `:app:testGithubDebugUnitTest` and `:app:lintGithubDebug` all pass (0 lint errors). `TrackerStatusLogic` and `TrackerFeedLogic` match the agreed rules exactly, including the two subtle ones: shared-IP precedence over whitelist state, and the company switch's `checked` derived from the resolved status rather than the raw whitelist key.

Four things worth fixing before merge.

### 1. The Monitored status is truncated in the default blocking mode

`item_tracker_feed_company.xml` caps `tvStatus` at `maxWidth="144dp"` with `maxLines="1"` and `ellipsize="end"`, but `statusString()` renders `MONITORED` as the full `tracker_monitored_minimal` — "Monitored — not blocked in Minimal mode" (39 chars, ~255dp at Material3 `LabelSmall`). It will clip to roughly "Monitored — not b…".

This is not an edge case: `BlockingMode.getDefaultMode()` returns `MODE_MINIMAL`, and any company known to TrackerControl's list but not to X-Ray/Disconnect resolves to `MONITORED`, so most rows in a default install look like this. It also squeezes `tvCompany` (weight-1, same row) down to ~113dp, so the company name ellipsizes as well.

Suggested fix: give `MONITORED` a short status word ("Monitored") and leave the full sentence to the expansion, which already shows it.

### 2. `getAppName()` does two PackageManager binder calls per bind

`TrackersListAdapter.getAppName()` calls `getApplicationInfo()` then `getApplicationLabel()`, and it is now invoked from `bindSection` (Essential explainer) and `bindCompany` (switch label) — i.e. inside `onBindViewHolder` while scrolling. Previously it only ran in the header bind. The plan claimed performance strictly improves; this is the one place it regresses. Caching it in a field is enough.

Related but minor: each company bind performs three or four separate `getDefaultSharedPreferences` round-trips (`isMinimalMode`, `isStrictMode`, and `isTrackerProtectionEnabled` twice — once directly, once inside `currentState`). Hoisting the mode booleans out of the per-row bind would be cheap.

### 3. A whitelisted category leaves a dead switch with no explanation

The old code kept the row tappable and showed the `category_unblocked_warning` toast. The new code resolves that state to `ALLOWED`/`READ_ONLY`, so `bindCompany` renders the "Allow X in Y" switch checked and disabled with nothing saying why. Replacing the toast was agreed; dropping the explanation was not. Reuse the existing string in `tvSharedIpNote` for that branch, the way `MONITORED` and `ALLOWED_SHARED_IP` already do.

### 4. Two smaller presentation points

- `ALLOWED_SHARED_IP` collapses to plain "Allowed", indistinguishable from a genuinely allowed company. The old row showed the "shared IP; enable Strict mode to block" caveat inline; it is now only visible after expanding. Defensible given the width constraint in (1), but it is a deliberate information loss worth confirming.
- The expand affordance is invisible: `item_tracker_feed_company.xml` has no chevron or caret, and the old `untap_unblock` tip died with the deleted layout, so nothing signals that rows are tappable.

### Verified clean

- No dangling references to the deleted layouts, `ExpandedListView` or `tvBlockingTip`.
- The header cards did get `layout_marginStart/End="16dp"` alongside `cardElevation="0dp"`, so removing the RecyclerView padding does not make them full-bleed.
- The `" *"` uncertainty legend is correctly re-homed into the per-company expansion via `tvUncertainNote` — more precise than the old per-category placement.
- The section switch keeps its `toggle_block_category_description` content description.
- Bind hygiene (null the listener before `setChecked`, reset every conditional visibility) is done on both switches.
- `ProtectionActivity`, `TrackerList`, `TrackerBlocklist` and `TrackersFragment` are untouched as planned; zero data migration, as intended.

One test nit: `noSubsetStateIsBlocked` passes `strict=true`, so it duplicates the last assertion of `sharedIpPrecedesWhitelistAndStrictOverridesIt` rather than covering the no-subset case it names.

PR #912, opened 2026-09-05
Fix easy tracker and native hotpath bugs

kasnder:
Fixes five independently reproduced hotpath findings:

- invalidate cached tracker verdicts when DNS evidence is refreshed
- reload policy when Internet blocking actually changes
- use 32-bit modular arithmetic for the native TCP send window
- forward zero-length UDP datagrams instead of treating them as EOF
- avoid redundant screen-off DoH eviction after cache hits

The audit document records these findings as fixed and leaves the harder findings open. Production-backed TCP-window and UDP-socket host regression tests are included; their Linux ASan/UBSan CI wiring remains follow-up work because the publishing credential cannot update workflow files.

Validation:

- focused Java/Robolectric policy and DNS tests
- existing native DNS-frame, IPv6-extension, DHCP-option, UDP-state and WireGuard flow-cache host suites under ASan/UBSan
- `assembleGithubDebug`, including C and Rust builds for all four Android ABIs
- `git diff --check origin/master...HEAD`


kasnder:
Superseded by the fixes now merged into master, including #924, #925 and combined Pixel-tested batch #926. The remaining #913 stack has been reconciled with master.

PR #917, opened 2026-09-05
Persist UDP redirect endpoints for each session

kasnder:
Java supplies a UDP redirect during the first policy decision, but the native session previously forgot it. The second datagram on the same tuple could therefore escape to the original destination, including on Secure DNS and direct-DNS routes.

Each admitted UDP session now stores the validated resolved address family, address and network-order port and uses it for every datagram. The original tuple remains the identity for lookup, response reconstruction and accounting.

Validation:
- production-backed UDP socket suite under ASan/UBSan
- repeated, unredirected, IPv4, IPv6, cross-family, invalid-address, original-tuple and response-tuple cases
- final `assembleGithubDebug` build

Stacked on #916.


kasnder:
UDP endpoint persistence and regression tests landed in the combined Pixel-tested batch #926. Closing as superseded.

